PR #23, opened 2016-09-09
opspec 0.1.2 support

chrisdostert:
fixes opspec-io/cli#8 (failure of serial operation run does not immediately fail all following operations)
fixes opspec-io/cli#15 (Docker for mac: Timeout exceeded while liveness checking engine)


chrisdostert:
approved

PR #87, opened 2017-02-18
fix) invalid value from environment results in infinite validation lo…

### LINKED ISSUE #81: Invalid input value from environment results in infinite validation loop
**setup**
1. create an op w/ an input constraint
2. set an environment var w/ same name as input and value not meeting defined constraint
3. run op

**expected**
display msg value from environment invalid & prompt for value

**actual**
validation error displayed over and over again in infinite loop

chrisdostert:
…op; closes #81


chrisdostert:
approved

PR #91, opened 2017-02-19
fix) Infinite validation loop when explicitly passing invalid value t…

### LINKED ISSUE #83: Child op input validation failure doesn't fail parent op
**setup**
1. create an op w/ no inputs
2. create a second op w/ an input constraint
3. reference the second op (child) from the first (parent)
4. run the first (parent) op

**expected**
both ops fail and input validation error displayed

**actual**
child fails but parent succeeds

### LINKED ISSUE #88: Infinite validation loop when passing a bad enum value
Tried to run the sample at: https://github.com/opspec-io/spec/blob/master/examples/azure/create-resource-group/op.yml with the following:

`opctl run -c azure/ -a location=asia create-resource-group`

chrisdostert:
fix) Infinite validation loop when explicitly passing invalid value to run cmd; closes #88
fix) Child op input validation failure doesn't fail parent op; closes #83

### DIFF
--- a/.opspec/test-coverage/docker-compose.yml
+++ b/.opspec/test-coverage/docker-compose.yml
@@ -50,4 +50,4 @@ services:
         exit 1
       fi
     environment:
-      - MINIMUM_COVERAGE=57
+      - MINIMUM_COVERAGE=59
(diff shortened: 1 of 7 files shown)


chrisdostert:
approved

chrisdostert:
approved

PR #93, opened 2017-02-19
fix) Node crash can leave the client hanging; closes #86

### LINKED ISSUE #86: Node crash can leave the client hanging
Presumably, the client should error out or do something other than appearing to work.

chrisdostert:


chrisdostert:
approved

chrisdostert:
approved

PR #97, opened 2017-02-22
feat) Automate node creation when using CLI closes #85

### LINKED ISSUE #85: Automate node creation when using CLI
User Story:
As a user of the CLI
I would like node creation to be automatic
In order save me from having to create a node

chrisdostert:


chrisdostert:
approved

PR #111, opened 2017-02-27
chore) Update to use bolt db for event storage; closes #110

### LINKED ISSUE #84: Output bound to env var not observed
**setup**
1. create an op
2. create a second op w/ an input, a container, and the input bound to an env var set in the container
3. reference the second op (child) from the first (parent)
4. create a third op w/ the output from the second op passed as an input, a container, and the input echo'd to stdout within the container
4. run the first (parent) op

**expected**
the env var value is printed

**actual**
nothing is printed because the env var has no value

### LINKED ISSUE #110: Update to use bolt db for event storage

chrisdostert:
fix) Output bound to env var not observed; closes #84

### DIFF
--- a/cli/version.go
+++ b/cli/version.go
@@ -2,5 +2,5 @@
 package main
 
 const (
-	version = "0.1.15-beta.118"
+	version = "0.1.15-alpha.41"
 )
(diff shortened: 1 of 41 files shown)


chrisdostert:
approved

PR #160, opened 2017-03-23
fix) Inputs need to be passed to child ops even though they are defau…

### LINKED ISSUE #158: Inputs need to be passed to child ops even though they are defaulted in the child ops
The parent op is forced to send inputs to child ops, even though there are defaults set.

chrisdostert:
…lted in the child ops; closes #158

### DIFF
--- a/cli/version.go
+++ b/cli/version.go
@@ -2,5 +2,5 @@
 package main
 
 const (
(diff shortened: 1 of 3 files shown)


chrisdostert:
approve

PR #170, opened 2017-03-28
chore) Don't recreate node on self-update; closes #169

### LINKED ISSUE #167: Multiple opctl networks created leading to lack of inter-container connectivity
It's currently possible to create multiple opctl networks leading to an unexpected lack of connectivity between containers

### LINKED ISSUE #169: Don't recreate node on self-update
self-update may be run under a different user (such as root in the common case sudo is used)

as such, to avoid surprises, we should not re-create the node at the end of the self-update call

chrisdostert:
fix) Multiple opctl networks created leading to lack of inter-container connectivity; closes #167


chrisdostert:
approve

PR #182, opened 2017-04-11
feat) Fail fast during parallel call; closes #154

### LINKED ISSUE #154: Fail fast during parallel call

chrisdostert:


chrisdostert:
approve

PR #189, opened 2017-04-15
feat) Support since in event filter; closes #187, fix) stdOut/stdErr …

### LINKED ISSUE #174: stdOut/stdErr output race condition
when chaining ops together and using stdOut/stdErr bindings as outputs, if the previous op ends to quickly the outputs do not always get picked up by the next op.

example of workaround:

sleep 5 echo "output1=foo" sleep 5

### LINKED ISSUE #187: Support "since" in event filter
Without a since attribute, any consumer of events may receive (potentially many) events it does not care about.

chrisdostert:
…output race condition; closes #174


chrisdostert:
approve

PR #193, opened 2017-04-26
feat) Add read from file as a new input source; closes #176

### LINKED ISSUE #176: Add read from file as a new input source
Similar to using a .env file to supply environment variable values.  
It would be great if we could optionally supply values via file.

This would allow fast swapping between environments by using different files  
And some usability improvements when an op has a substantial amount of inputs.

E.g. Input sources:
 - arg provided via -a option
 - **.env file** <- New option
 - env var
 - default
 - prompt

chrisdostert:


chrisdostert:
approve

PR #204, opened 2017-05-03
fix) Errors in parallel block don't fail op; closes #201

### LINKED ISSUE #201: Errors in parallel block don't fail op
**expected**:
if an error occurs during execution of a container or op in a parallel block the parent op fails

**actual**:
if an error occurs during execution of a container or op in a parallel block the parent op succeeds

chrisdostert:


chrisdostert:
approve

PR #229, opened 2017-05-19
fix) Op Fails w/ Msg='io: read/write on closed pipe'; closes #217

### LINKED ISSUE #217: Op Fails w/ Msg='io: read/write on closed pipe'
Running the debug op on the  beta build errors out with **Msg='io: read/write on closed pipe'**

chrisdostert:


chrisdostert:
approve

PR #230, opened 2017-05-20
fix) Not providing sub op param causes panic; closes #227

### LINKED ISSUE #227: Not providing sub op param causes panic

chrisdostert:


chrisdostert:
approve

PR #242, opened 2017-06-01
fix) Node crash if docker image doesn't exist; closes #241

### LINKED ISSUE #241: Node crash if docker image doesn't exist
Attempting to run a containerCall op with an image that doesn't exist causes the node to crash. 

e.g.: node:6-apline

You just see `Event channel closed unexpectedly`

chrisdostert:


chrisdostert:
approve

PR #271, opened 2017-07-17
feat) number/string/object => container file binding support

### LINKED ISSUE #226: when using args.yml if the yml is invalid the opc does not notify

chrisdostert:
feat) Object param support
feat) Error if args.yml is invalid; closes #226


chrisdostert:
approve

PR #347, opened 2017-10-15
feat) automatic host detection

chrisdostert:


chrisdostert:
@habibbhutto this should allow you to use any host; not just localhost

PR #445, opened 2018-04-07
chore) publish explicit versions to docker hub

chrisdostert:


cli/version.go
```diff
@@ -2,5 +2,5 @@
 package main
 
 const (
-	version = "0.1.24"
+	version = "0.1.24-alpha.159"
```
SudhindraKovalam: was this intentional ?

chrisdostert: @SudhindraKovalam thanks, always the careful eye : ) . It was just because I ran a test release; it doesn't have any effect on anything. This file is actually go source code so the released binary will know it's own version. It's what get's returned from `opctl -v` for example.

PR #632, opened 2019-09-28
move cli/{core,util} to cli/internal

chrisdostert:


apexskier:
Why does this add so many lines of code if it's a simple move (as the commit title indicates)?

chrisdostert:
> Why does this add so many lines of code if it's a simple move (as the commit title indicates)?

move includes some refactoring

Issue #659, opened 2019-12-09
Piping opctl output makes op exit non-gracefully


apexskier:
I tend to run my op like this: `opctl run dev | jqs` (`jqs` is equivalent to `jq  -Rr ". as \$line | try (fromjson | .) catch \$line"`) (this somewhat works around #62 and #76).

For a minimal reproduction, you can pipe your task to `| tee`.

When I `ctl+c` this piped command, opctl doesn't exit gracefully, which means all my containers are still running and I can't restart it due to port conflicts.

Related to https://github.com/opctl/opctl/issues/657

apexskier:
This just bit me again. Without #657, there's no way to find the tasks to kill them.

PR #693, opened 2020-02-17
Gracefully exit when SIGTERM is received

apexskier:
This allows graceful cleanup when piping opctl output into other commands.

Resolves https://github.com/opctl/opctl/issues/659

### DIFF
--- a/cli/internal/core/runer.go
+++ b/cli/internal/core/runer.go
@@ -98,13 +98,19 @@ func (ivkr _runer) Run(
 	)
 
 	// init signal channel
-	intSignalsReceived := 0
-	signalChannel := make(chan os.Signal, 1)
-	defer close(signalChannel)
-
+	sigIntsReceived := 0
+	sigIntChannel := make(chan os.Signal, 1)
+	defer close(sigIntChannel)
 	signal.Notify(
-		signalChannel,
-		syscall.SIGINT, //handle SIGINTs
+		sigIntChannel,
+		syscall.SIGINT,
+	)
+
+	sigTermChannel := make(chan os.Signal, 1)
+	defer close(sigTermChannel)
+	signal.Notify(
+		sigTermChannel,
+		syscall.SIGTERM,
 	)
 
 	nodeHandle, createNodeIfNotExistsErr := ivkr.nodeProvider.CreateNodeIfNotExists()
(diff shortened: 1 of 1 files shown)


apexskier:
I'll need assistance if maintaining coverage is required. My code is similar to the existing SIGINT handling. Both could use tests, but I wasn't able to figure out how to test that sort of behavior in go.

cli/internal/core/runer.go
```diff
@@ -146,10 +152,9 @@ func (ivkr _runer) Run(
 	for {
 		select {
 
-		case <-signalChannel:
-			if intSignalsReceived == 0 {
-
-				intSignalsReceived++
+		case <-sigIntChannel:
+			sigIntsReceived++
```
chrisdostert: Thx for contribution! @apexskier why is sigIntsReceived needed?

chrisdostert:
> I'll need assistance if maintaining coverage is required. My code is similar to the existing SIGINT handling. Both could use tests, but I wasn't able to figure out how to test that sort of behavior in go.

@apexskier personally I think it makes sense to maintain coverage and have unit test over this; it's pretty critical path. 

Grr @golang for not exposing standard libs via interfaces to enable unit testing. The only way to unit test it would be to introduce an abstraction exposed via interface w/ a fake. This is what I had started https://github.com/golang-interfaces for since anyone will have the same issues but it's a PITA : / . 

Personally I think it warrants it.. I invited you to that org if you wanted to go that route.

chrisdostert:
you can see an example here of this approach w/ os package and testing exit was called w/ certain exit code https://github.com/opctl/opctl/blob/9e8c29fd3b0b228210670aff2166417762c0f0eb/cli/internal/cliexiter/cliExiter.go#L38

cli/internal/core/runer.go
apexskier: It's a rename of `intSignalsReceived`, which exists to support a second ctrl+c forcefully stopping execution.

I renamed because it's not used for the new signal handler. I changed the logic slightly here to ensure it's actually counting. Previously it was essentially a boolean flag (which I actually might change it to) since it was only incrementing once.

apexskier:
> personally I think it makes sense to maintain coverage and have unit test over this; it's pretty critical path.

I'll make another attempt. I'll call out that the existing path doesn't have unit tests; since this fixes a bug I feel like merging without unit tests is worthwhile.

cli/internal/core/runer.go
```diff
@@ -98,13 +98,19 @@ func (ivkr _runer) Run(
 	)
```
apexskier: another note - the fact that this filename isn't spelled right is real annoying (runner, not runer)

apexskier:
Took a couple hours to try to test via this mechanism - https://stackoverflow.com/a/33404435, but because the test framework itself handles these signals. I'm not able to allocate more time into this now.

chrisdostert:
thanks for trying and for the contribution. I agree, we can get the tests in separately.

cli/internal/core/runer.go
chrisdostert: contribution welcome ; )

PR #694, opened 2020-02-17
Opctl docker image without dind

lmarszal:
The goal of this PR is to enable following: WHEN hosting opctl node in docker container THEN use same docker engine for spinning up opctl containers.
Currently the blocker for that was host path mapping - from path relative to container to absolute to docker host.
This changes unblocks that.

Proven to work:
```
description: dev
name: dev
inputs:
  appDir:
    dir:
      default: .
  dockerSocket:
    socket:
      description: |
        Must be the correct socket path to the host machine's dockerd runtime socket IE /var/run/docker.sock (default
        path). For linux, you can try netstat -lx | grep docker | grep -Po '/.*docker.sock' to verify the proper socket
        file.
run:
  serial:
    # prepare env
    - container:
        image: { ref: "alpine" }
        dirs:
          /app: $(appDir)
        cmd:
          - sh
          - -ce
          - |
            touch file
            rm -f opctl/pid.lock
            rm -f copy
            mkdir -p opctl
        workDir: /app
    # run test
    - parallel:
      - container:
          name: opctl
          image: { ref: "opctl/opctl:0.1.26-cli.alpha.3" }
          dirs:
            /app: $(appDir)
            /root/opctl: $(appDir/opctl)
          sockets:
            /var/run/docker.sock: dockerSocket
      - container:
          image: { ref: "opctl/opctl:0.1.26-cli.alpha.3" }
          envVars:
            OPCTL_CLI_API_BASEURL: http://opctl:42224/api
          dirs:
            /app: $(appDir)
            /root/opctl: $(appDir/opctl)
          cmd:
            - sh
            - -ce
            - |
              # wait for server to start
              sleep 3
              # copy file
              opctl run -a src=/app/file -a dst=/app -a dstPath=copy github.com/opspec-pkgs/fs.cp-file#2.0.1
```


.opspec/release/to-docker-cli/op.yml
```diff
@@ -0,0 +1,55 @@
+name: release/to-docker-cli
```
chrisdostert: @lmarszal how do you feel about adding a check in the entrypoint.sh of the existing Dockerfile which looks for `/var/run/docker.sock` and if it exists it doesn't start a docker daemon and instead uses it? Everything else is same right?

chrisdostert: basically make this part optional based on existence of docker socket mounted at conventional path https://github.com/opctl/opctl/blob/master/.opspec/release/to-docker/entrypoint.sh#L3-L23

lmarszal: I did thought about that. My thinking was to make it look like `docker` image on docker hub.
Docker has two sets of images:
 * `docker:latest` - that has only cli
 * `docker:dind` - that has dockerd and requires privileged mode
I was thinking about the same pattern here (`opctl` having didn and `opctl:cli` having only cli).
But if we don't see benefits in shipping opctl image with cli only your solution sounds better.

chrisdostert: either seems reasonable. I was just debating a little on additional maintenance of 2 images but no major concerns. Pick your poison ; )

PR #702, opened 2020-02-28
Omit SCGContainerCallImage.Ref if empty

apexskier:
This fixes a bug if a opspec file is generated using `src` instead of `ref` for a container.

I found this when working on https://github.com/Remitly/infra-ops/tree/master/fpkg/run, looks like @psamaan may have manually patched this in vendored code 😬?

### DIFF
--- a/sdks/go/model/scg.go
+++ b/sdks/go/model/scg.go
@@ -32,7 +32,7 @@ type SCGContainerCall struct {
 
 type SCGContainerCallImage struct {
 	Src       *string       `json:"src,omitempty"`
-	Ref       string        `json:"ref"`
+	Ref       string        `json:"ref,omitempty"`
 	PullCreds *SCGPullCreds `json:"pullCreds,omitempty"`
 }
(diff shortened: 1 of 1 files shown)


sdks/go/model/scg.go
```diff
@@ -32,7 +32,7 @@ type SCGContainerCall struct {
 
 type SCGContainerCallImage struct {
 	Src       *string       `json:"src,omitempty"`
-	Ref       string        `json:"ref"`
+	Ref       string        `json:"ref,omitempty"`
```
psamaan: I think Ref's type also needs to be a pointer so that it can be an empty value. See https://stackoverflow.com/questions/38486564/unmarshal-marshal-json-with-int-set-to-0-does-not-seem-to-work

chrisdostert:
@apexskier  thanks. I have a WIP branch i'm making that a pointer now that it's nil'able but this is a quick win in the meantime.

PR #844, opened 2020-12-17
Merge go modules to support IDE development

apexskier:
One of the most problematic parts of working in this codebase is the split between SDK and CLI. Because they're set up as individual go modules, opening this repo in most IDEs (vscode, intellij) doesn't resolve the go project successfully, and intellisense and code navigation is not available (at least without manual configuration). This also causes a lot of duplication in vendor directories (~1.2 million lines of extra code), which slows things down a lot due to indexing.

My workaround so far has been to open the sdks/go and cli directories individually, which leads to a lot of jumping back and forth, and makes it hard to globally find things.

In this PR, I've merged the two go modules into one at the root level, which addresses these issues and allows the project to "just work" when I open in my IDE

intellisense between the go sdk and the cli


apexskier:
I don't know why the automated build is failing. Running the cli `opcl run test` works on my computer. Running the full root `opctl run build` fails with an out of memory error building the website.

<details>

```
The build failed because the process exited too early. This probably means the system ran out of memory or someone called `kill -9` on the process.
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
ContainerExited Id='9d4a575fefa14ada8e0aa14bc999beda' ImageRef='docker.io/library/node:12.8.1-alpine' Outcome='FAILED' Error='nonzero container exit code. Exit code was: 1' Timestamp='2020-12-17T12:41:52Z'

Error='nonzero container exit code. Exit code was: 1' Id='f4b896035ad048a6b9afc1f05de27d1d' OpRef='/Users/cameronlittle/Developer/opctl/webapp/.opspec/build' Timestamp='2020-12-17T12:41:52Z'
```

</details>

apexskier:
I think it's build flakiness. A different error is now showing up.

chrisdostert:
> I think it's build flakiness. A different error is now showing up.

: /

PR #845, opened 2020-12-17
Remove CLI exiter in favor of go error propagation, allow cleanup on errors

apexskier:
This is a fairly decent size refactor to remove the current pattern of direct `os.Exit` calls from internal cli code in favor of propagating an error back up the call chain. This is more idiomatic go code, and should **increase reliability** of the cli by allowing deferred functions to execute when errors occur instead of terminating immediately. It also avoids needing to have weird commented return statements in the main code to support tests and allows testing error behavior more explicitly.

Basically, all the main cli command interface methods now return an error instead of terminating program exit. The top level cli then wraps these calls in a function that checks the error and outputs appropriate messaging and status codes. The cli exiter interface has been removed.

I've also hoisted the cli colorer out of any internal CLI code - that all now relies on the cli outputer interface, which should make opctl respect color disabling in a few more cases.

Another important reason to do this is to respect the `mow.cli`'s instruction to call it's Exit function instead of os.Exit - https://github.com/jawher/mow.cli#exiting


chrisdostert:
Thanks! The main reason for the exiter interface was for unit testing. This PR drops codecov by 5.5%; it would be desirable to at least maintain what we had coverage wise.

apexskier:
I've added some more tests and have brought that number up, but overall I think the code coverage diff is a little misleading. If you look at most of the code quality metrics, the code is effectively tested the same amount, but because the line count for each exit call is now less, the percentage goes down (3 lines of code that were tested previously are now 1 line of code).

For example, `cli/internal/core/op/validater.go` shows a -30% reduction in test coverage. Previously, the lines in red in this screenshot were tested, but since the code has been simplified the apparent amount of code coverage diff has gone down because there's other code in the file that hasn't been tested.
Screen Shot 2020-12-18 at 11 19 30

In my opinion, this is a sign that a lot of the existing test coverage is testing structural abstraction layers, rather than actual business logic.

> The main reason for the exiter interface was for unit testing.

If you look at the change, the pattern of "exiting" from a function is still tested. It's a lot smoother now; you can directly check the return value of these functions, instead of needing to set up a mock interface and verify how it's called.

chrisdostert (approved):
Thanks! LGTM

PR #851, opened 2021-01-05
fix release/to-docker binary path

chrisdostert:


apexskier (commented):
It looks like there are a bunch of unrelated changes to the JS SDK? What are those?

It also looks like the file `opctl.linux` has been accidentally committed.

chrisdostert:
@apexskier 
re: unrelated changes) yeah those should have probably gone in separate PR but I started scratching an itch. We upgraded to react 17 a PR or two ago and I was just doing some other semi-related package upgrades to touch that up. 
re: `opctl.linux`) yeah good call we should clean that up in next PR.

PR #853, opened 2021-01-05
Reorganize ops and run full test suite on build

apexskier:
The main impetus for this is the fact that tests are currently broken on main. This is because SDK tests, webapp tests, and api validation are not run during the main build that PR checks use.

Additionally, this consolidates a few ops to the top level opspec. Now that there is a single go project, this makes more sense to me.

Note that the code coverage stats may change significantly as the SDK code is now included in coverage.


chrisdostert:
hmm, this makes the go SDK different from all the other SDK's in that it no longer has a .opspec. For me it conceptually doesn't make sense to move the ops out of the SDK project from the standpoint this is a mono-repo and the SDK's each are their own project separate from the other projects like the CLI, WebApp, etc..

chrisdostert:
At the very least, could we maybe break this into 2 PR's; 1 for running all tests in all projects on PR (which I think is great), and 1 for moving the go SDK ops?

apexskier:
> For me it conceptually doesn't make sense to move the ops out of the SDK project from the standpoint this is a mono-repo and the SDK's each are their own project separate from the other projects like the CLI, WebApp, etc..

For me, it conceptually makes more sense to think of this as a single go project, which also aligns with #844. It's still multiple projects, but from a go perspective you are never going to make changes in one project without wanting to verify functionality across all projects. This also aligns with community standards around [go project layout](https://github.com/golang-standards/project-layout) (although this project currently uses different directory names, `cli` ≈ `cmd/cli`, `sdks/go` ≈ `pkg`).

However, I think it's fine to leave a test op in the go sdk directory to align with the other sdks.

apexskier:
I'm going to get rid of the build op in the sdk directory though, as it's just running tests and reporting code coverage. Code coverage reporting can only happen once, and the test op will remain if you want to use it. (As described, build doesn't make sense since there are no "vetted artifact(s)" produced).

apexskier:
The currently broken test is fixed in #864

.opspec/generate/op.yml
```diff
@@ -4,17 +4,16 @@ inputs:
   path:
     string:
       description: relative path to file containing a 'go:generate ...' pre-processing directive
-      default: ./cli/...
+      default: ./...
```
chrisdostert, line 7: What's the motivation for moving this out of cli project? It *seems* appropriately placed in the cli project. There is also one in the sdks/go project which *seems* appropriate as well. cli and sdk/go are separate projects; I might be working in one and want to `opctl run generate`; don't want to cd to root of repo to do that. Just my 2cents.

chrisdostert (approved):

.opspec/generate/op.yml
apexskier: It feels to me like there should be a root one, given that's where the go mod is. I can add one for the cli, no big deal.

PR #854, opened 2021-01-05
Fix docker container removal

apexskier:
#736 changed the name opctl assigns to docker containers, but it didn't update the teardown code, which can result in a docker container remaining alive after opctl stops tracking it. If ports are bound, this will cause the next op run to fail.


chrisdostert:
Thanks!

chrisdostert (approved):

PR #857, opened 2021-01-08
Delete accidentally committed binary

apexskier:
Also, update the gitignore to prevent this. I've left the cli line in the gitignore in case people have old binaries hanging around.


chrisdostert (approved):
Thanks!

PR #859, opened 2021-01-10
Emit CallStarted events for skipped conditional branches

apexskier:
Currently, a call ended event is emitted when a skipped if conditional branch call is processed without a start event. This means that anything watching the events stream will receive an ended event without a corresponding start event, and won't know how to deal with it.

I've introduced a new outcome type: "skipped", to indicate when this happens. Alternatively, failed if conditional could be not outputted at all, but that would mean anything listening to the event stream wouldn't know about them at all, which limits the user experience.

### DIFF
--- a/cli/internal/clioutput/cliOutput.go
+++ b/cli/internal/clioutput/cliOutput.go
@@ -152,7 +152,7 @@ func (this _cliOutput) containerExited(event *model.Event) {
 	switch event.CallEnded.Outcome {
 	case model.OpOutcomeSucceeded:
 		this.Success(message)
-	case model.OpOutcomeKilled, model.OpOutcomeSkipped:
+	case model.OpOutcomeKilled:
 		this.Info(message)
 	default:
 		this.Error(message)
@@ -200,7 +200,7 @@ func (this _cliOutput) opEnded(event *model.Event) {
 	switch event.CallEnded.Outcome {
 	case model.OpOutcomeSucceeded:
 		this.Success(message)
-	case model.OpOutcomeKilled, model.OpOutcomeSkipped:
+	case model.OpOutcomeKilled:
 		this.Info(message)
 	default:
 		this.Error(message)
--- a/cli/internal/clioutput/cliOutput.go
+++ b/cli/internal/clioutput/cliOutput.go
@@ -152,7 +152,7 @@ func (this _cliOutput) containerExited(event *model.Event) {
 	switch event.CallEnded.Outcome {
 	case model.OpOutcomeSucceeded:
 		this.Success(message)
-	case model.OpOutcomeKilled:
+	case model.OpOutcomeKilled, model.OpOutcomeSkipped:
 		this.Info(message)
 	default:
 		this.Error(message)
@@ -200,7 +200,7 @@ func (this _cliOutput) opEnded(event *model.Event) {
 	switch event.CallEnded.Outcome {
 	case model.OpOutcomeSucceeded:
 		this.Success(message)
-	case model.OpOutcomeKilled:
+	case model.OpOutcomeKilled, model.OpOutcomeSkipped:
 		this.Info(message)
 	default:
 		this.Error(message)
--- a/cli/internal/core/runer.go
+++ b/cli/internal/core/runer.go
@@ -192,17 +192,17 @@ func (ivkr _runer) Run(
 
 			ivkr.cliOutput.Event(&event)
 
-			if nil != event.CallEnded {
-				if event.CallEnded.Call.ID == rootCallID {
-					switch event.CallEnded.Outcome {
-					case model.OpOutcomeSucceeded, model.OpOutcomeSkipped:
-						return nil
-					case model.OpOutcomeKilled:
-						return &RunError{ExitCode: 137}
-					default:
-						return &RunError{ExitCode: 1}
-					}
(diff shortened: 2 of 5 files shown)


apexskier:
Tests are hanging, and I don't know why. They hang on different cases locally and in the build check.

apexskier:
Locally, running the unit tests I added work

chrisdostert:
> Tests are hanging, and I don't know why. They hang on different cases locally and in the build check.

Looking at gh actions workflow it looks like e2e tests are hanging so speaking from experience, some code here is probably broken.

chrisdostert:
> > Tests are hanging, and I don't know why. They hang on different cases locally and in the build check.
> 
> Looking at gh actions workflow it looks like e2e tests are hanging so speaking from experience, some code here is probably broken.

We're using sharness in the e2e tests. I've found it semi difficult to determine which test(s) are hanging in sharness because it seems to not print test output until after the test completes (as opposed to acknowledging a given test has started when it starts). There might be some way to get sharness to acknowledge start of test but I haven't looked into it that far. What i've done in the past is to just disable blocks of tests until we get a pass and then start re-enabling until we find the test that's hanging. Since these changes are fairly specific i'm guessing something in the introduced changes results in some set of calls never resulting in a call end event being received.

chrisdostert:
> > > Tests are hanging, and I don't know why. They hang on different cases locally and in the build check.
> > 
> > 
> > Looking at gh actions workflow it looks like e2e tests are hanging so speaking from experience, some code here is probably broken.
> 
> We're using sharness in the e2e tests. I've found it semi difficult to determine which test(s) are hanging in sharness because it seems to not print test output until after the test completes (as opposed to acknowledging a given test has started when it starts). There might be some way to get sharness to acknowledge start of test but I haven't looked into it that far. What i've done in the past is to just disable blocks of tests until we get a pass and then start re-enabling until we find the test that's hanging. Since these changes are fairly specific i'm guessing something in the introduced changes results in some set of calls never resulting in a call end event being received.

alternatively we might be able to configure sharness with a test timeout and get it to log what test is timing out

apexskier:
It looks like it was transient - they worked on the latest build

chrisdostert:
Side Note) I'd love to replace sharness with an opctl parallelLoop but to do that kind of testing we'd need some capability to handle non-zero exit codes. Right now opctl assumes any container exiting non-zero is a failure and fails fast.

chrisdostert:
All the above aside, i'm thinking about the actual PR and have questions. Right now I believe the intention is for no CallStarted event to be fired if a call is skipped and so no CallEnded event would be fired either. In my opinion this makes sense because the call is never started, it was skipped. 

Q1) What are you trying to do that explicitly needs to know the call was skipped
Q2) If something does truly need to explicitly know a call was skipped (which seems reasonable but we haven't needed it yet), it seems like maybe we should introduce a new event type CallSkipped.

apexskier:
> Q1) What are you trying to do that explicitly needs to know the call was skipped
Display a call graph UI

> Q2) If something does truly need to explicitly know a call was skipped (which seems reasonable but we haven't needed it yet), it seems like maybe we should introduce a new event type CallSkipped.
That seems reasonable as well. The main issues is mismatched start/end events.

chrisdostert:
> That seems reasonable as well. The main issues is mismatched start/end events.

Agreed. So to me it seems like behavior we're arriving at is: 
If call is skipped, Emit CallSkipped event. No CallStarted or CallEnded events should be emitted for said call.

chrisdostert:
Wait, can't the client watch for CallStarted event and check the If property to see if it was skipped?

apexskier:
> Wait, can't the client watch for [CallEnded](https://github.com/opctl/opctl/blob/86fdeb329a5881868a1fa8e8476fd31041238296/sdks/go/model/event.go#L34) event and check the [Call.If](https://github.com/opctl/opctl/blob/86fdeb329a5881868a1fa8e8476fd31041238296/sdks/go/model/call.go#L15) property to see it was skipped?

Yes, that's another option, but it doesn't really make semantic sense to me.

apexskier:
The root issue of the tests hanging is that the root `Call`'s errors are swallowed here.

https://github.com/opctl/opctl/blob/86fdeb329a5881868a1fa8e8476fd31041238296/sdks/go/node/core/startOp.go#L78

Previously, an ended event was emitted, but now it's not. (this doesn't happen on my branch since I've restructured the error pipeline, but that's a way larger change than the scope of this PR)

apexskier:
I think I'm just going to close this for now and depend on a call ended without a skip for now to avoid the difference right now.

chrisdostert:
> Yes, that's another option, but it doesn't really make semantic sense to me.

I hear you. I could see a case being made either way and be convinced. On one hand you could look at it as "the call should never start because it's skipped". On the other hand you could look at it as "how could it end if it was never started". 

Now that I've gotten back into the headspace I slightly recall thinking about this before and I think there's also a pragmatic element. The end event was sufficient to convey all the information needed so we only sent that. If you think of a parallelLoop or something and 100k elements being skipped, that's 100k extra events being transmitted/stored etc.. that aren't explicitly needed by anything. That said, it could be argued it's a pre-optimization etc...  and semantics are more important. 

I think changing the code to emit a CallStarted event in the case of skip would be reasonable but I do see tradeoffs. 

I'm totally on board kicking the can on this for another day though so +1 to closing w/out changes (for now).

PR #860, opened 2021-01-10
Add additional vscode debugging features + native toolchain instructions

apexskier:
The first new vscode launch configuration allows attaching to a built cli binary launched directly from the host, instead of through the dev op.

The second allows running a given test file with an attached debugger. This lets you avoid modifying the test op to quickly debug a given test file. Since this project uses Ginko for testing, you don't have the option to use the vscode go plugin's auto-test detection, so this is a decent alternative.

I've also added instructions for development using native go tooling instead of dev ops. This is valuable for people more familiar with the go ecosystem, and it's quite a bit faster than relying on the ops on macOS due to how docker interacts with the filesystem.


CONTRIBUTING.md
```diff
@@ -59,3 +59,23 @@ Pull requests are subject to:
 
 - approval by one or more [maintainers](https://github.com/orgs/opctl/teams/maintainers/members)
 - the [build](.opspec/build) op continuing to run with a successful outcome
+
+# Local development with native go tooling
+
+First, ensure dependencies are installed
+
+- [go 1.15](https://golang.org/doc/go1.15)
```
chrisdostert, line 67: is it reasonable to refer to the compile op for go version (or alternatively store go version somewhere and reference that in all the ops) to keep a single source of truth? Otherwise I think it might get hard/error prone to track and update all the places we're one off documenting versions of things

chrisdostert (approved):

CONTRIBUTING.md
apexskier: I can refer to the version in the go mod file, as that's the main thing go uses - there's about 9 different ops that refer to the go version.

chrisdostert: I like that. In the ops we could parse that file and grab version but seems overkill.. something to consider anyways

PR #861, opened 2021-01-12
port interpretation draft

malizz:


sdks/go/opspec/interpreter/call/container/ports/interpret.go
```diff
@@ -0,0 +1,29 @@
...
+
+// Interpret container ports
+func Interpret(
+	scope map[string]*model.Value,
+	containerCallSpecPorts map[string]string,
+	scratchDirPath string,
+) (map[string]string, error) {
+	containerCallPorts := map[string]string{}
+	for callSpecContainerPortNumber, callSpecContainerPortExpression := range containerCallSpecPorts {
+		// @TODO: use reference.interpret once reference syntax no longer optional
```
apexskier, line 18: Can you avoid this TODO given this is a new type, so it doesn't have backwards compatibility?

sdks/go/opspec/interpreter/call/container/ports/pkg.go
```diff
@@ -0,0 +1,4 @@
+// Package ports exposes functionality for interpreting ports of container calls.
+package ports
+
+//go:generate go run github.com/maxbrunsfeld/counterfeiter/v6 -generate
```
apexskier, line 4: What's this fake used for?

chrisdostert:
Closing due to inactivity. Please re-open if you can spare time to get this functionality in. That would be awesome!

PR #862, opened 2021-01-13
Restructure to align api client and core interfaces

apexskier:
This PR, inspired by work done in my `mini-opctl` branch, aligns the APIClient and Core objects to a common core node interface. The intent is to to align error and context cancellation patterns, to simplify testing, and (long term) allow consumers like the CLI to choose whether they'd like to use a remote opctl node or run ops directly.

I've added some additional context plumbing and error returning where one of the prior interfaces defined it, but won't fully plumb this info like we should long term in this PR.

I've removed the `cli/internal/model` package in favor of defining the interfaces within it in the places that consume them (`NodeCreateOpts` into the local node provider package, `RunOpts` in the node runner, and the node handle into the node provider package). This makes it more obvious how these are used, and given they all fit very well into their new homes I prefer it.

I've removed the `<-chan error` type in the pubsub and local node ("core") event streams. Even though I think it makes sense to have a way to return multiple errors, in practice the returned error channel was _always ignored_, and was only used for websocket errors, not opctl-specific ones. For now, I think the downsides of removing this outweigh the upsides of aligned interfaces. I want to follow up with better error handling in both the api and local nodes.

The main reasons for the increase in line count is that there are three generated fakes where there were previously two: APIClient, local node "Core", and the new shared OpNode; and the fact that I had to add a _lot_ of new tests to fulfill the `codecov/patch` check.


cli/cli.go
```diff
@@ -106,21 +104,17 @@ func newCli(
 	}
 
 	cli.Command("auth", "Manage auth for OCI image registries", func(authCmd *mow.Cmd) {
+		authCmd.Command("add", "Add auth for an OCI image registry", func(addCmd *mow.Cmd) {
```
apexskier, line 107: Most line changes here just make indentation consistent between commands.

cli/cli.go
```diff
@@ -70,15 +68,15 @@ func newCli(
 			Value:  "127.0.0.1:42224",
 		},
 	)
-	nodeCreateOpts := model.NodeCreateOpts{
+	nodeCreateOpts := local.NodeCreateOpts{
```
apexskier, line 71: You'll notice that this is now more obviously coupled to a "local" node. Since this type wouldn't be used with a remote one, I think this makes a lot more sense.

cli/internal/core/auth/add.go
```diff
@@ -36,12 +36,7 @@ func (ivkr _adder) Add(
 	username string,
 	password string,
 ) error {
-	nodeHandle, err := ivkr.nodeProvider.CreateNodeIfNotExists()
```
apexskier, line 39: This is now handled by the lazylocalnode wrapper when using an api client based OpNode.

chrisdostert (commented):
This is a nit but it would be great to observe the test pattern outlined in [the contribution guidelines](https://github.com/opctl/opctl/blob/main/cli/CONTRIBUTING.md#contribution-guidelines) as far as arrange/act/assert pattern & objectUnderTest naming convention.

apexskier:
I updated the tests that work in that pattern. Most of the ones I added aren't behavioral, they're just ensuring code execution works through the layers of abstraction, and I only added them to pass the `codecov/patch` check. (It's a bit of a code smell to me to have layers of abstraction like this that don't wrap behavior, but that's a separate discussion)

chrisdostert:
Thanks! I'm a fan of the consolidation of interfaces. I was going to push some more changes to this branch to consolidate even further but I figure let's just get this in and i'll do that separately. As much as possible wanna avoid long running massive PR's.

chrisdostert (approved):

Issue #856, opened 2021-01-05
CallKiller test is flaky


apexskier:
## 🐛 Bug Report

Running SDK tests produces failures on the `main` branch sometimes.

## To Reproduce

```
opctl run sdks/go/.opspec/test
```

## Expected behavior

Passing tests 100% of the time

## Actual Behavior

`callKiller` fails in a github build - https://github.com/opctl/opctl/pull/853/checks?check_run_id=1650813815

<details>

<summary>weird op failure locally</summary>

```
▻ opctl run sdks/go/.opspec/test
OpStarted Id='67955b0d26b54eba9bbb987bb497e7a7' OpRef='/Users/cameronlittle/Developer/opctl/sdks/go/.opspec/test' Timestamp='2021-01-05T17:12:40Z'

ContainerStarted Id='97924a0c86464000b7e80a6febf0c487' OpRef='/Users/cameronlittle/Developer/opctl/sdks/go/.opspec/test' ImageRef='docker.io/library/golang:1.15' Timestamp='2021-01-05T17:12:40Z'

1.15: Pulling from library/golang
Digest: sha256:c47d6c4ee25e34c97f3b1a92923a913e266a2b4b95731e8f5aa9065ca9e7c7fc
Status: Image is up to date for golang:1.15
unit testing code
sdks/go/node/core/containerCaller.go:10:2: readdirent: input/output error
ContainerExited Id='97924a0c86464000b7e80a6febf0c487' ImageRef='docker.io/library/golang:1.15' Outcome='FAILED' Error='nonzero container exit code. Exit code was: 1' Timestamp='2021-01-05T17:12:49Z'

Error='nonzero container exit code. Exit code was: 1' Id='46c2173fd6154a1fb834dba57c4d2415' OpRef='/Users/cameronlittle/Developer/opctl/sdks/go/.opspec/test' Timestamp='2021-01-05T17:12:49Z'

OpEnded Id='67955b0d26b54eba9bbb987bb497e7a7' OpRef='/Users/cameronlittle/Developer/opctl/sdks/go/.opspec/test' Outcome='FAILED' Error='nonzero container exit code. Exit code was: 1' Timestamp='2021-01-05T17:12:49Z'


1 ↩
```

</details>

<details>

<summary>callKiller fails locally</summary>

```
▻ opctl run sdks/go/.opspec/test
[...]

PR #863, opened 2021-01-13
Ensure UI uses the correct listen address

apexskier:
Currently, the UI uses a hardcoded `localhost:42224`, instead of the provided cli argument listen address.

### DIFF
--- a/cli/cli.go
+++ b/cli/cli.go
@@ -318,7 +318,8 @@ func newCli(
 				"Opctl web UI opened!",
 				ui(
 					cliParamSatisfier,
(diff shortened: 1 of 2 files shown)


apexskier:
#856 is causing the build failure

chrisdostert (approved):
Thanks! I think this looks nice but is it reasonable to merge this after #869? I think it will be net easier that way. Alot of the CLI code touched here got deleted or moved into the cli.go file in that PR so won't apply anymore. That PR also fixes #856. I'm happy to update this PR after.

PR #864, opened 2021-01-13
Fix race condition in call killer

apexskier:
This fixes the root cause of the flaky test discovered in #853. After updating the test to show more data, I was seeing failures like this:

```
[1b2c75a5 ./sdks/go/.opspec/test]       Expected
[1b2c75a5 ./sdks/go/.opspec/test]           <[]string | len:3, cap:3>: ["dummyNode3ID", "dummyNode3ID", "dummyNode2ID"]
[1b2c75a5 ./sdks/go/.opspec/test]       to contain element matching
[1b2c75a5 ./sdks/go/.opspec/test]           <string>: dummyNode1ID
[1b2c75a5 ./sdks/go/.opspec/test] 
```

I'm not sure exactly why that was happening (there _could_ be an issue with the pubsub?), but it indicates a race.

A simpler fix is to remove the goroutine waiting. Publishing already uses goroutines internally to avoid blocking, so this was unnecessary complexity, and it doesn't match how publishing is used elsewhere.

### DIFF
--- a/sdks/go/node/core/callKiller.go
+++ b/sdks/go/node/core/callKiller.go
@@ -2,9 +2,6 @@ package core
 
 import (
 	"context"
-	"fmt"
-	"runtime/debug"
-	"sync"
 	"time"
 
 	"github.com/opctl/opctl/sdks/go/model"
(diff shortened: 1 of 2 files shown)


chrisdostert:
Thanks! I *think* this was a miss when it was last refactored. It previously was the case that we would actually perform the kill in the go routine vs publishing the kill request and since that was sync did each in a go routine so the kills happened in parallel.

chrisdostert (approved):

PR #867, opened 2021-01-14
Pass context throughout call paths where possible

apexskier:
This sets a better pattern for cleanup after op completion or error and should overall help with reliability.

There are now only three places where we create a "root" context (background only, no TODOs)

1. When initializing the CLI
2. When killing containers. This can be triggered when the root context is cancelled, so we don't want to use that previously cancelled context.
3. When the API handler starts an op. This is an asynchronous API, and the op keeps running after the initial start request finishes.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -29,6 +29,21 @@
             "request": "launch",
             "mode": "test",
             "program": "${file}"
+        },
+        {
+            "name": "Run and debug an opctl node",
+            "type": "go",
+            "request": "launch",
+            "mode": "debug",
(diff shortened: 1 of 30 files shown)


apexskier:
#856 is causing the build failure

chrisdostert (approved):
Thanks! I think these look nice but is it reasonable to merge this after https://github.com/opctl/opctl/pull/869? I think it will be net easier that way. Alot of the CLI code touched here got deleted or moved into the cli.go file in that PR so won't apply anymore.  That PR also fixes #856. I'm happy to update this PR after.

PR #868, opened 2021-01-19
Investigating CallKiller flakiness 

apexskier:


sdks/go/pubsub/pubSub.go
```diff
@@ -62,48 +62,43 @@ func (ps *pubSub) Subscribe(
...
+		err := ps.eventStore.List(ctx, filter, dstEventChannel)
+		if err != nil {
+			return
 		}
 
 		// new events
 		for event := range publishEventChannel {
 			select {
-			case dstEventChannel <- event:
 			case <-ctx.Done():
```
apexskier, line 91: Optimistically cancel, like other things do

sdks/go/pubsub/pubSub.go
```diff
@@ -62,48 +62,43 @@ func (ps *pubSub) Subscribe(
 	<-chan model.Event,
 	error,
 ) {
-	dstEventChannel := make(chan model.Event, 1000)
```
apexskier, line 65: I don't know why these'd be buffered, and 1000 feels arbitrary.

sdks/go/pubsub/pubSub.go
```diff
@@ -62,48 +62,43 @@ func (ps *pubSub) Subscribe(
...
+	dstEventChannel := make(chan model.Event)
 
 	go func() {
 		defer close(dstEventChannel)
 
-		publishEventChannel := make(chan model.Event, 1000)
-		defer ps.gcSubscription(publishEventChannel)
+		publishEventChannel := make(chan model.Event)
+		defer ps.closeSubscription(publishEventChannel)
```
apexskier, line 71: Name change for understandability. Garbage collection isn't guarunteed

sdks/go/pubsub/eventStore.go
```diff
@@ -60,47 +58,42 @@ func (es *_eventStore) Add(
...
-			defer it.Close()
-			sinceBytes := []byte(es.eventsByTimestampKeyPrefix + sinceTime.Format(sortableRFC3339Nano))
-			for it.Seek(sinceBytes); it.ValidForPrefix([]byte(es.eventsByTimestampKeyPrefix)); it.Next() {
-				item := it.Item()
-				item.Value(func(v []byte) error {
-					event := model.Event{}
-					if err := json.Unmarshal(v, &event); nil != err {
-						return err
-					}
+	eventChannel chan model.Event,
```
apexskier, line 61: This has been simplified to be blocking, and write directly to the channel. It's called from a concurrent context that mirrors that channel back anyways, so this is a nice simplification.

sdks/go/pubsub/eventStore.go
```diff
@@ -60,47 +58,42 @@ func (es *_eventStore) Add(
...
-						case <-ctx.Done():
-							return ctx.Err()
-						case eventChannel <- event:
-						}
+		it := txn.NewIterator(badger.DefaultIteratorOptions)
+		defer it.Close()
+
+		sinceBytes := []byte(es.eventsByTimestampKeyPrefix + sinceTime.Format(sortableRFC3339Nano))
+		for it.Seek(sinceBytes); it.ValidForPrefix([]byte(es.eventsByTimestampKeyPrefix)); it.Next() {
+			if err := it.Item().Value(func(v []byte) error {
```
apexskier, line 74: This error was being swallowed

chrisdostert:
Just FYI I solved this in https://github.com/opctl/opctl/pull/869 but you should still carry on w/ what you're doing in here. The reason for the failure seemed to be the test was expecting exactly once delivery but our pubsub is at least once delivery.

chrisdostert:
@apexskier this one's been hanging around a bit; I feel like there's some valuable stuff in here but I know some of it has been covered elsewhere. Any update/thoughts?

apexskier:
> Any update/thoughts?

nope. As you can see tests are failing and I haven't figured out why.

chrisdostert:
@apexskier any time to get this in? It seems like some of the things in here are useful. Otherwise we should probably close close to avoid dead PR building up.

chrisdostert:
I put these in #987 so will close this one

PR #873, opened 2021-02-28
Stop using equinox.io; closes #866

chrisdostert:

PR #874, opened 2021-02-28
Stop vendoring

chrisdostert:


apexskier (approved):
I'd prefer this to be split out from https://github.com/opctl/opctl/pull/873 (currently both commits are here) but I'm definitely in favor of doing this.

chrisdostert:
> I'd prefer this to be split out from #873 (currently both commits are here) but I'm definitely in favor of doing this.

For sure; I'll rebase before I merge; just wanted to avoid conflicts so had it as my upstream

Issue #808, opened 2020-10-24
Stop using Fake pubsub in tests; use real pubsub. 


chrisdostert:
## 💥 Proposal

Opctl is a distributed, event driven architecture. We currently fake/mock pubsub in tests wherever it's a dependency. The proposal here is to stop faking/mocking it and instead use the real thing. 

This will give us better tests that match real world behavior, not fake mocked out behavior. 

See [this example](https://github.com/opctl/opctl/blob/6dc7a064bc9b49253398a913aa1cccca40df986e/sdks/go/node/core/caller_test.go#L139) of where/how we fake it.

PR #876, opened 2021-03-02
Remove some unnecessary 3p deps; Further work on #808

chrisdostert:


sdks/go/data/fs/handle.go
```diff
@@ -53,7 +48,7 @@ func (lh handle) rListDescendants(
 	[]*model.DirEntry,
 	error,
 ) {
-	childFileInfos, err := lh.ioUtil.ReadDir(path)
+	childFileInfos, err := ioutil.ReadDir(path)
```
apexskier, line 51: Just an FYI if you hadn't seen, `io/ioutil` is being deprecated in go 1.16. This'll be moved to `os.ReadDir`

apexskier (approved):
We need to get gofmt PR checks added, there's some violations being added here. I'm not commenting on them explicitly since it's so easily automatable though and is just noise.

apexskier:
> We need to get gofmt PR checks added, there's some violations being added here. I'm not commenting on them explicitly since it's so easily automatable though and is just noise.

https://github.com/opctl/opctl/pull/877

sdks/go/data/fs/handle.go
chrisdostert: Nice. I gave a quick shot at upgrading to 1.16 but tests weren't passing so whoever gets to it first will have to look into it; assuming won't be too bad. We should ditch `io/ioutil` when we do.

chrisdostert:
> > We need to get gofmt PR checks added, there's some violations being added here. I'm not commenting on them explicitly since it's so easily automatable though and is just noise.
> 
> #877

awesome

PR #877, opened 2021-03-03
Add test check that all go source has been formatted

apexskier:
Here's what it looks like when this fails
Check that all go source has been formatted · opctlopctl@343bf76 - Google Chrome - 2021-03-03 at 11 50 46


sdks/go/tools.go
```diff
@@ -1,11 +1,11 @@
 // +build tools
 
-package go
+package go_sdk_tools
```
apexskier, line 3: `go` is a keyword, so it's not a valid identifier name (at least according to `gofmt`)

chrisdostert (approved):
This is great

apexskier:
weird error here in `serialCaller`

PR #884, opened 2021-03-12
Use unique temporary directory in tests instead of global one

apexskier:
This makes one-off testing more reliable during local development. I've run into a few cases where running local tests, e.g. through vscode's debugger, fails due to leftover files from prior test runs.

### DIFF
--- a/cli/internal/nodeprovider/local/createNodeIfNotExists_test.go
+++ b/cli/internal/nodeprovider/local/createNodeIfNotExists_test.go
@@ -12,7 +12,9 @@ var _ = Context("CreateNodeIfNotExists", func() {
 	It("shouldn't panic", func() {
 		/* arrange */
 		dataDir, err := ioutil.TempDir("", "")
-		Expect(err).To(BeNil())
+		if err != nil {
+			panic(err)
+		}
 		nodeProvider := New(NodeCreateOpts{
 			DataDir: dataDir,
 		})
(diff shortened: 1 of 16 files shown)


cli/internal/nodeprovider/local/createNodeIfNotExists_test.go
```diff
@@ -2,15 +2,19 @@ package local
...
 	. "github.com/onsi/ginkgo"
-	"os"
+	. "github.com/onsi/gomega"
 )
 
 var _ = Context("CreateNodeIfNotExists", func() {
 	It("shouldn't panic", func() {
 		/* arrange */
+		dataDir, err := ioutil.TempDir("", "")
+		Expect(err).To(BeNil())
```
chrisdostert, line 15: Not a big deal but you might consider panic'ing if these fail rather than failing the test. Thinking being we're not testing if ioutil.TempDir errs; if it does then we can't test so panic seems appropriate.

chrisdostert:
Nice!

chrisdostert (approved):

cli/internal/nodeprovider/local/createNodeIfNotExists_test.go
apexskier: Sure, that's fine. I wasn't quite sure what the preferred pattern should be.

chrisdostert:
@apexskier Heads up; I don't know why but codecov check seems flickery recently. I found if it doesn't publish on here then I can run test locally and check completes..

PR #886, opened 2021-03-14
Embed local file/dir values as objects when calling remote node

chrisdostert:
Today when I start an op on a remote node and I want to pass it a local file or directory as an arg, I have to write code to embed the files/dirs as objects. I would like whatever mechanism is needed to pass local files/dirs as args to an op to be performed automatically so I don't have to do it myself and so that as future enhancements like content defined chunking (CDC) and dedupe  are exposed via our data API, they will automatically be leveraged.

### DIFF
--- a/sdks/go/node/api/client/embedLocalFilesAndDirs.go
+++ b/sdks/go/node/api/client/embedLocalFilesAndDirs.go
@@ -1,0 +1,96 @@
+package client
+
+import (
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+
+	"github.com/opctl/opctl/sdks/go/model"
+)
+
+var oneMB int64 = 1024 * 1024
+var maxEmbedBytes int64 = 40 * oneMB
+
+func getMaxEmbedMB() float64 {
+	return float64(maxEmbedBytes) / float64(oneMB)
+}
+
+//embedLocalFilesAndDirs mutates args by embedding any referenced file &/or dirs as objects.
+// This makes the args location independent and therefore runnable on a remote node.
+// note1: this loads referenced files/dirs into memory so we limit max combined embed to maxEmbedBytes to reduce the chances of memory exhaustion.
+// note2: this approach is primitive; we will need to support de-dupe and chunking in the future.
+func embedLocalFilesAndDirs(
+	args map[string]*model.Value,
+) error {
+	var totalSize int64
+	for key, val := range args {
+		var fileOrDirPath string
+		if nil != val.Dir {
+			fileOrDirPath = *val.Dir
+		} else if nil != val.File {
+			fileOrDirPath = *val.File
+		} else {
+			continue
+		}
+
+		fileObj, size, err := fileOrDirPathToObject(fileOrDirPath)
+		if nil != err {
+			return err
+		}
+
+		totalSize += size
+		if totalSize > maxEmbedBytes {
+			return fmt.Errorf("embedding failed: combined size of files/dirs cannot exceed %gMb", getMaxEmbedMB())
+		}
+
+		args[key] = &model.Value{Object: &fileObj}
+	}
+	return nil
+}
+
+// fileOrDirPathToObject recursively serializes a file or directory to an object
+func fileOrDirPathToObject(
+	path string,
+) (map[string]interface{}, int64, error) {
+	info, err := os.Stat(path)
+	if nil != err {
+		return nil, 0, err
+	}
+
+	if !info.IsDir() {
+		if info.Size() >= maxEmbedBytes {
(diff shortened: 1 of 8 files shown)


sdks/go/node/api/client/embedLocalFilesAndDirs.go
```diff
@@ -0,0 +1,73 @@
...
+	"os"
+	"path/filepath"
+)
+
+//embedLocalFilesAndDirs mutates args by embedding any referenced file &/or dirs as objects.
+// This makes the args location independent and therefore runnable on a remote node.
+// note1: this loads referenced files/dirs into memory so can easily exhaust available memory.
+// note2: this approach is primitive; we will need to support de-dupe and chunking in the future.
+func embedLocalFilesAndDirs(
+	args map[string]*model.Value,
```
tallen94, line 16: does this need to be a ptr, since this func doesnt return anything?
```
args *map[string]*model.Value
```

chrisdostert: no because maps in golang are reference types

sdks/go/node/api/client/embedLocalFilesAndDirs.go
```diff
@@ -0,0 +1,73 @@
...
+func fileOrDirPathToObject(
+	path string,
+) (map[string]interface{}, error) {
+	info, err := os.Stat(path)
+	if nil != err {
+		return nil, err
+	}
+
+	if !info.IsDir() {
+		body, err := ioutil.ReadFile(path)
```
apexskier, line 48: I think it would be worth adding a max size check here (over the entire recursive operation), given the known potentially memory issues and the potential negative impact on the remote node if it gets sent a large amount of information.

apexskier:
I'd like to see a little more context as to why we're doing this in the PR description, for future historical reasons.

sdks/go/node/api/client/embedLocalFilesAndDirs.go
chrisdostert: The max seems arbitrary here since the remote node will have arbitrary ram. Do you have more details about your thoughts? This is basically meant to be an incremental step towards the proper solution which would leverage content defined chunking (CDC) and dedupe via further enhancements to our data API. In practice the code here is used today by at least one client I know of and so getting it into the SDK so that future enhancements can be centrally rolled out was the main goal.

chrisdostert:
> I'd like to see a little more context as to why we're doing this in the PR description, for future historical reasons.

Done! LMK if that doesn't cover it

sdks/go/node/api/client/embedLocalFilesAndDirs.go
apexskier: This happens locally, though, right?

Just add a check that we're not reading more than, say, 200 mb and bail if that happens. It'll catch obvious issues like someone trying to mount their whole disk or a giant cache tree or something.

chrisdostert: The thing is; I know clients doing this today that in practice are sending upwards of 2GB.

chrisdostert: I *hope* this won't be a thing for long. It's a primitive temporary solution but since it's the only possible solution today, it's what clients are doing. We need to do more work on the universal data API to support doing this the right way. I think if we set it to 2GB it would suffice for now; it feels arbitrary but I'm onboard if you'd like to see it.

sdks/go/node/api/client/embedLocalFilesAndDirs.go
tallen94: gotcha

sdks/go/node/api/handler/handler.go
```diff
@@ -41,6 +41,9 @@ func (hdlr _handler) ServeHTTP(
 	httpResp http.ResponseWriter,
 	httpReq *http.Request,
 ) {
+	// limit req size to 1 Gb
+	httpReq.Body = http.MaxBytesReader(httpResp, httpReq.Body, 1000000000)
```
chrisdostert, line 45: @apexskier req size limiting added here

apexskier: This limits it on the remote only, right? I'm more concerned about memory explosions on the users local computer.

chrisdostert: I missunderstood your ask; I was thinking because it's in the clients hands they would check that if they cared but I guess being an SDK it could be nice to do it for them. Should be good to go now https://github.com/opctl/opctl/pull/886/files#diff-77477245ac34983a608813751c412801f40e453bcce7a691930704d74bcf61dbR13

chrisdostert: Also, I reduced max size down to 40mb. I think until we have a more sophisticated implementation it's reasonable for us to be overly restrictive. The case I know of where folks were sending up to 2Gb can be handled by the user.

apexskier (approved):

PR #887, opened 2021-03-15
Update UI to use universal data API

chrisdostert:


webapp/src/constructDataUrl.ts
```diff
@@ -1,7 +1,5 @@
-import path from 'path'
-
 export default function constructDataUrl(
     dataRef: string
 ): string {
-    return `http://localhost:42224/api/pkgs/${encodeURIComponent(path.dirname(dataRef))}/contents/${encodeURIComponent(path.basename(dataRef))}`
+    return `http://localhost:42224/api/data/${encodeURIComponent(dataRef)}`
```
apexskier, line 4: Why doesn't this route show up in `sdks/go/node/api/urlTemplates.go`?

apexskier (commented):
Can you add some background information to the PR description? I hadn't seen the "universal data API" this is referring to. Does anything else use it? Is the endpoint it's replacing (which looks deprecated) still used once this is merged?

webapp/src/constructDataUrl.ts
chrisdostert: it's in the API spec and is implemented but the path wasn't put into that file. It probably should be added.

chrisdostert: It's meant to be the replacement to the `pkgs/{ref}` and `pkgs/{ref}/contents/{path}` endpoints.

apexskier (approved):

PR #889, opened 2021-03-15
Use local mock server for testing git pull errors

apexskier:
This moves away from relying on real network connections, which speeds up tests locally (especially from overseas) and makes tests more robust. It also avoids some issues I was seeing where cloudfront is preventing me from running tests.

Unfortunately, I wasn't able to find a way to create a mock http handler that acts as a remote git server (the protocol is a little hard to follow). This would allow completely fixing the issues I'm having.


chrisdostert (approved):
I know this get's into opinion territory, but I feel whenever possible, testing real behavior as opposed to mocked/fake behavior, is preferable because it's the only way to have true confidence. In practice I've experienced many occasions where real dependency behavior differed from mock/faked behavior and your tests lose value.  

That said, I realize these tests haven't been reliable on your internet connection. I'm wondering if instead of the mocking/faking we might do something like tag these tests with RequiresGoodInternet or something so we could allow you and others with troublesome internet to readily exclude them and keep productivity up. 

LMK what you think.

apexskier:
I think it makes full sense to make this change. 

> In practice I've experienced many occasions where real dependency behavior differed from mock/faked behavior and your tests lose value.

This is true, but the defined dependency isn't really well defined. That's explicitly obvious in the 403 handler, which previously had a note on why it couldn't use GitHub vs gitlab, so we already know there are problems. This switches to testing the actual intended behavior.

apexskier:
@chrisdostert I don't have permissions to merge this, with the codecov not passing.

chrisdostert:
It thinks project coverage went down; project coverage is the one (so far) we've decided to require not to go down in a PR

chrisdostert:
> @chrisdostert I don't have permissions to merge this, with the codecov not passing.

It thinks project coverage went down; project coverage is the one (so far) we've decided to require not to go down in a PR

chrisdostert:
> the defined dependency isn't really well defined. That's explicitly obvious in the 403 handler, which previously had a note on why it couldn't use GitHub vs gitlab, so we already know there are problems.

I already approved this so again; I'm not blocking the PR or anything but there is a difference of opinion and that's ok. To me the fact we had to use two providers to test the intended scenarios is not a sign of anything other than different providers have different behavior and we chose them according to what we wanted to test.

Issue #798, opened 2020-08-17
Opspec validation for bad `output` names


apexskier:
## 🚀 Feature

Opctl should fail when an invalid `output` name is used.

## Motivation

Given this op:

```yml
description: Run the app
name: run
run:
  serial:
    - op:
        ref: github.com/opspec-pkgs/object.merge#1.0.0
        inputs:
          default:
            a: 1
            b: 2
            c: 3
          override:
            b: asdf
        outputs:
          vars:
    - container:
        image: { ref: 'alpine' }
        envVars: $(vars)
        cmd: ['env']
```

I get this error message:
```
unable to interpret '$(vars)' as envVars; error was unable to interpret $(vars) to object; error was unable to interpret 'vars' as reference; 'vars' not in scope
```

## Pitch

The error message currently points to my subsequent usage in the container as the error (where I use "envVars"). This is misleading, as it's actually the `run.serial[0].op.outputs` that contains an error. This ideally would be a validation check that would show me the opspec that's incorrect.

PR #893, opened 2021-03-19
Detect invalid outputs in opspecs

apexskier:
This introduces validation of outputs when making op calls. Currently, if an invalid output is used, it's ignored. When used later in an op, it's easy to misinterpret the errors produced as happening in the wrong place.

I made an attempt to try to detect common user-errors and proactively provide information to help solve the error (see test cases).

I also consolidated the scattered logic for converting a variable reference to its name (`$(foo)` to `foo`) into a single, unit tested, utility function.

This is technically backwards incompatible - currently if an invalid op output is specified and never used, it will be silently ignored. This changes so it will now cause an error, as it's not a valid reference. Resolving is as simple as removing the bad output line, so I think this is definitely worth the increased usability, given the v0 status of opctl.

Resolves https://github.com/opctl/opctl/issues/798

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -68,14 +68,24 @@ First, ensure dependencies are installed
 - [gpgme](https://www.gnupg.org/related_software/gpgme/) (on macOS, `brew install gpgme`)
 - [dlv](https://github.com/go-delve/delve) (for debugging) - Because this project uses go modules, install this globally with `go get github.com/go-delve/delve` to run the tool outside of the project directory.
 
+## CLI debugging
+
 Build a binary with
 
 `go build -o opctl-beta ./cli`
 
-To debug
+Run with [dlv](https://github.com/go-delve/delve), e.g.
 
 `go run github.com/go-delve/delve/cmd/dlv --check-go-version=false --listen=127.0.0.1:40000 --headless=true --api-version=2 exec /PATH/TO/OPCTL/opctl-beta run dev`
 
-This will suspend execution until a client connects on port 40000. If using VSCode, you can use the saved run configuration to connect.
+This will suspend execution until a client connects on port 40000. If using VSCode, you can use the "Connect to dlv debugger" run configuration to connect.
 
 On macOS, you can debug exit signal handling by sending an interrupt to the underlying PID. When the debugger connects, it will give you the with a message like: `Got a connection, launched process ../opctl/opctl-beta (pid = 79924).` (path to process and PID will be different). Kill with `kill -SIGINT 72958`.
+
+## Node debugging
+
+Ops are run in a "node" running as an external process, which means debugging the CLI directly doesn't catch most of the logic. To debug this external process, run `opctl node kill` to kill the current node (attached to port 42224), then use the VSCode "Run and debug an opctl node" run configuration to spin up a new node with the debugger attached or debug the CLI command "opctl node create".
+
+## Tests
+
+This project uses the [Ginkgo](https://github.com/onsi/ginkgo) test framework, which means some IDE's go tooling won't support normal methods of running tests. Run the VSCode "Run and debug this test suite" run configuration while viewing a test file to run the tests attached to the debugger.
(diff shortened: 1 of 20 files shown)


chrisdostert:
Awesome! Semi-related; another thing I've been noodling on is requiring variables to be defined before being used. Without that you don't know if a variable binding is supposed to define or assign which easily can result in gotchas. I need to create a feature proposal for it.

CONTRIBUTING.md
```diff
@@ -68,14 +68,24 @@ First, ensure dependencies are installed
 - [gpgme](https://www.gnupg.org/related_software/gpgme/) (on macOS, `brew install gpgme`)
 - [dlv](https://github.com/go-delve/delve) (for debugging) - Because this project uses go modules, install this globally with `go get github.com/go-delve/delve` to run the tool outside of the project directory.
 
+## CLI debugging
+
 Build a binary with
 
 `go build -o opctl-beta ./cli`
```
chrisdostert, line 75: the cli/.opspec/compile op completes pretty fast these days; I wonder if it's worth pointing to use that since it might make for fewer moving parts.

sdks/go/opspec/ref_test.go
```diff
@@ -0,0 +1,19 @@
+package opspec
+
+import (
+	"testing"
+
+	. "github.com/onsi/gomega"
+)
+
+func TestRefToName(t *testing.T) {
```
chrisdostert, line 9: Just a nit but for consistency might be nice to have wire up same as all the other tests; I also retroactively noticed this variance in https://github.com/opctl/opctl/blob/c5611ff2884eef74c2531064ec8776bea0eaf739/sdks/go/internal/aggregate_error/aggregate_error_test.go#L12 which I was actually hoping to cleanup. It's one of those things I'd hope we just do one way; the way we choose doesn't seem super important (unless it is to you); moreso just find value in there being a single way used throughout.

chrisdostert (approved):

sdks/go/opspec/interpreter/call/op/outputs/interpret.go
```diff
@@ -1,28 +1,73 @@
...
 	opPath string,
 	opScratchDir string,
 ) (
 	map[string]*model.Value,
 	error,
 ) {
-
 	outputArgs, err := params.Coerce(outputArgs, outputParams, opScratchDir)
-	if nil != err {
+	if err != nil {
```
chrisdostert, line 25: I think we should do a mass replacement of this ordering. Some languages you can accidentally assign the thing you're comparing so putting the constant on the left keeps you from doing that accidentally. That's not a thing in golang and the ecosystem is more used to err on left so I think I should concede and go with the flow.

chrisdostert: https://github.com/opctl/opctl/pull/897

chrisdostert:
@apexskier also just a reminder you'll want to update CHANGELOG.md

CONTRIBUTING.md
apexskier: That's fine, just a note that this is specifically under the section "# Local development with native go tooling"

sdks/go/opspec/ref_test.go
apexskier: It's a far nicer IDE experience to use standard go test patterns - I can debug a single test without requiring running the entire suite.

sdks/go/opspec/interpreter/call/op/outputs/interpret.go
apexskier: I very much agree. I see warnings about this from my default linters sometimes (https://staticcheck.io/docs/checks#ST10170) and the `nil != err` pattern isn't something I've seen in any other go codebase.

sdks/go/opspec/ref_test.go
apexskier: ref_test go — opctl - Code - 2021-03-22 at 10 37 27

apexskier: I had to add the [VSCode launch task](https://github.com/opctl/opctl/blob/6b4e1974c70f80c561110ab873d506ef734b7c62/.vscode/launch.json#L27) to do test debugging because of Ginkgo, and it's a lot more overhead since it runs all tests in the suite.

PR #902, opened 2021-03-20
Allow defining description on call graph nodes; closes #900

### LINKED ISSUE #615: vscode intellisense error
## Repro
1. https://opctl.io/docs/setup#vscode
1. create a minimal valid `op.yml`

## Expected
1. no red squiggleys

## Actual
1. red squiggleys on first character of file due to vscode not being able to fully interpret the schema. If u hover over it u see:
image

## Possible Causes
1. vscode yaml plugin can't handle patternProperties 
2. the way we merge all our schemas into one schema isn't generating a proper schema see [opspec/opfile/.opspec/build](https://github.com/opctl/opctl/tree/master/opspec/opfile/.opspec/build)

### LINKED ISSUE #900: Allow defining description on call graph nodes
## 🚀 Feature
Currently you can define a `name` for calls but it would be nice to also support `description`.

### existing (assuming https://github.com/opctl/opctl/issues/898)
```yaml
inputs:
  domain:
    description: txt to echo
    string: {}
run:
  name: echo
  container:
    image:
      ref: alpine
    cmd: [echo, $(txt)]
```

### proposed (assuming https://github.com/opctl/opctl/issues/898)
```yaml
inputs:
  - name: domain
    description: txt to echo
    string: {}
run:
  name: echo
  description: This container will take the input $(txt) and log it via the echo comman
  container:
    image:
      ref: alpine
    cmd: [echo, $(txt)]
```

chrisdostert:
Dont generate jsonschema.json; Define schema as single file; closes #615

Notes re jsonschema.json file consolidation: it seems like the jsonschema generated by [json-schema-ref-parser](https://www.npmjs.com/package/json-schema-ref-parser) is too complicated (or possibly invalid) for general jsonschema client consumption. Putting everything back into a single file and using standard JSON schema $ref's seems to solve the problem and arguably makes things more standard.


apexskier:
> Dont generate jsonschema.json; Define schema as single file; closes #615

It looks like this won't be resolved with this PR (testing locally). I'm guessing that won't be fixed until https://github.com/SchemaStore/schemastore/blob/master/src/schemas/json/opspec-io-0.1.7.json is updated.

op yml — opctl - Code - 2021-03-22 at 11 38 04

sdks/go/model/opSpec.go
```diff
@@ -13,6 +13,7 @@ type OpSpec struct {
 //CallSpec is a spec for a node of a call graph; see https://en.wikipedia.org/wiki/Call_graph
 type CallSpec struct {
 	Container    *ContainerCallSpec    `json:"container,omitempty"`
+	Description  string                `json:"description,omitempty"`
```
apexskier, line 16: Why isn't this a `*string` reference, like `Name`?

sdks/go/model/param.go
```diff
@@ -2,66 +2,74 @@ package model
...
-	String  *StringParam  `json:"string,omitempty"`
+	Array       *ArrayParam   `json:"array,omitempty"`
+	Boolean     *BooleanParam `json:"boolean,omitempty"`
+	Description string        `json:"description,omitempty"`
+	Dir         *DirParam     `json:"dir,omitempty"`
+	File        *FileParam    `json:"file,omitempty"`
+	Number      *NumberParam  `json:"number,omitempty"`
+	Object      *ObjectParam  `json:"object,omitempty"`
+	Socket      *SocketParam  `json:"socket,omitempty"`
+	String      *StringParam  `json:"string,omitempty"`
```
apexskier, line 13:
```suggestion
	Description string        `json:"description,omitempty"`
	
	// only one of the following is expected
	Array       *ArrayParam   `json:"array,omitempty"`
	Boolean     *BooleanParam `json:"boolean,omitempty"`
	Dir         *DirParam     `json:"dir,omitempty"`
	File        *FileParam    `json:"file,omitempty"`
	Number      *NumberParam  `json:"number,omitempty"`
	Object      *ObjectParam  `json:"object,omitempty"`
	Socket      *SocketParam  `json:"socket,omitempty"`
	String      *StringParam  `json:"string,omitempty"`
```

apexskier (approved):
This PRs a little confusing, since there are two things going on here:

1. The title refers to one thing, #900, which is to add support for a description field on opsec calls and opspec inputs. This sense to me.
2. The opening comment refers to another, #615, which is to restructure the json schema to fix VSCode parsing. Also makes sense

Combining the two things in a single PR makes it very hard to tell what actually changed with the JSON schema for the first item, since the entire json schema file was updated. It would be nice to split the two into two PRs (one made to the other branch) so I can review each change independently (two commits would also work, given the rebase enforcement in the codebase).

chrisdostert:
> > Dont generate jsonschema.json; Define schema as single file; closes #615
> 
> It looks like this won't be resolved with this PR (testing locally). I'm guessing that won't be fixed until https://github.com/SchemaStore/schemastore/blob/master/src/schemas/json/opspec-io-0.1.7.json is updated.
> 
> op yml — opctl - Code - 2021-03-22 at 11 38 04

If you're relying on the schemastore for sure  we also need to submit a PR with this update (we need to submit a PR anyways because the one they have is old). If you're relying on the file like is mentioned in https://opctl.io/docs/setup/bare-metal#vscode then you'll need to update link. This did seem to resolve it for me though in my testing.

sdks/go/model/opSpec.go
chrisdostert: Just because it was `string` [on parameter already](https://github.com/opctl/opctl/blob/8a41a4b8da60b8a34ea1426e004462622c62ad53/sdks/go/model/param.go#L27) and for our use cases differentiating `nil` vs empty isn't meaningful. `Name` probably should be same but it already exists.

chrisdostert:
> This PRs a little confusing, since there are two things going on here:
> 
> 1. The title refers to one thing, #900, which is to add support for a description field on opsec calls and opspec inputs. This sense to me.
> 2. The opening comment refers to another, #615, which is to restructure the json schema to fix VSCode parsing. Also makes sense
> 
> Combining the two things in a single PR makes it very hard to tell what actually changed with the JSON schema for the first item, since the entire json schema file was updated. It would be nice to split the two into two PRs (one made to the other branch) so I can review each change independently (two commits would also work, given the rebase enforcement in the codebase).

The title got auto-generated by github from the PR message. I started out adding description property but I actually got blocked by re-generating the schema. It seemed I couldn't generate a valid one anymore. So that led to me needing to fix the schema generation to solve the original task. In this case it was a really minimal schema generation change because it was just adding a single string attribute to an existing object.

Issue #909, opened 2021-03-31
Serial loop using until doesn't seem to work


apexskier:
## 🐛 Bug Report

An example op using serial loop and until doesn't work.

## To Reproduce

```yaml
name: test
description: |
  This is an example op.yml file that counts down using a serial loop
inputs:
  counter:
    number:
      default: 2
run:
  serialLoop:
    until:
      - eq: [$(counter), 0]
    run:
      container:
        image: { ref: alpine }
        cmd:
          - sh
          - -c
          - |
            echo $(counter)
            echo "\$(($(counter)-1))" > /file
        files:
          /file: $(counter)
```

## Expected behavior

Emits 2, 1, 0, then exits.

## Actual Behavior

Emits 2, 1, 0, -1, -2, ... and doesn't exit.

If I switch `eq` with `ne`, the op exits immediately, so the predicate is being evaluated. Adding `if: [eq: [$(counter), 2]` behaves as expected (the op runs), so eq is broken for numbers.

chrisdostert:
This too me awhile to debug but it turns out It's your echo; you need to add `-n` or else it will include a newline at the end and "0\n" isn't equal to 0

chrisdostert:
Site note) Having to start a container to do math is not so great; we should add native math operators to opspec.

PR #910, opened 2021-03-31
Update opspec reference documentation

apexskier:
This is a major refresh of the website's opspec documentation at https://opctl.io/docs/reference/opspec/index, adapted from my `mini-opctl` branch. The overall goal is to make learning how to write ops easier.

Some of the major themes:

I've softened the language used within these docs to feel less jarringly technical (removing shorthands like `\w` and `&`, removing `[{type}]` from links and titles). I've added a link to the JSON schema for those users who want to view the strict schema opspecs require.

I've added more examples, both in code form and with explanations of why and how various features might be used. I went into much more detail about things like scoping, needs, looping, and predicates, which I think is a big improvement - it's currently very hard to understand how these work without copy-pasting and messing around yourself.

I've reorganized the structure to be less deeply nested and more explorable. The "Opspec > Op > Op" section in the sidebar has been collapsed to "Opspec > op.yml". The "conceptual" and shared items within "Opspec > Op > Op > Call" have been hoisted up into the "op.yml" layer, which I think makes the "Call" section a lot more understandable - it's just the specific types of calls - and makes the shared items more discoverable (predicates, for example).

Within pages describing object types, I've removed the table of contents of properties from the body, in favor of the existing table of contents provided by docasaurus.

This documentation reflects my current understanding of how ops work (I might have gotten something wrong).


website/docs/reference/cli/index.md
```diff
@@ -0,0 +1,6 @@
+---
+title: CLI
+sidebar_label: Overview
+---
+
+The opctl cli provides command line access to using ops and managing a node.
```
apexskier, line 6: There is a broken link on https://opctl.io/docs/introduction, which doesn't have an appropriate page to link to. This is a quick and dirty fix (I don't want to introduce significant changes to the CLI docs in this PR).

chrisdostert: Good catch; this was probably my miss when I updated CLI docs to be hierarchical vs a single page. I wish [linking to categories was a thing in docusaurus](https://github.com/facebook/docusaurus/issues/2643); hopefully it gets in soon.

website/docs/reference/opspec/op.yml/predicate.md
```diff
@@ -0,0 +1,72 @@
+---
+title: Predicate
+---
+
+A predicate is a condition that evaluates to a boolean value, used for conditional logic. The predicate is an object with a single key defining the type of condition.
+
+## `eq`
+
+An [array](../types/array). When all items in the array are equivalent, the predicate is true.
```
apexskier, line 9: @chrisdostert Is type coercion applied during `eq`/`ne` predicate evaluation? (related to https://github.com/opctl/opctl/issues/909)

bbrooks:
@apexskier it's hard to navigate through this much content in a git diff. Is it possible to host this change so it can be explored in the same way as https://opctl.io/docs/reference/opspec/index/? That way we could get a sense of how it would be to explore these new docs organically.

apexskier:
Good idea, but branch-hosted builds are not currently a thing within this codebase, and I don't have an obvious place to put them right now.

If you want to explore them, you can clone the repo, run `opctl run dev` from the `website` directory, and visit `localhost:3000`.

apexskier:
I messed up codecov by uploading a coverage file from an old branch. No go file changes though, so there's no change

website/docs/reference/opspec/op.yml/index.md
```diff
@@ -0,0 +1,49 @@
+---
+sidebar_label: Overview
+title: op.yml
+---
+
+The `op.yml` file is a [YAML 1.2](https://yaml.org/spec/1.2/spec.html) file. It defines an operation's metadata, inputs, outputs, and call graph.
+
+The `op.yml` file must conform to the `opspec` [JSON schema](https://github.com/opctl/opctl/blob/main/opspec/opfile/jsonschema.json). Opctl will interpret your `op.yml` dynamically and may produce runtime errors for additional reason such as invalid variables or variable coercion.
+
+## Properties
```
tallen94, line 10: Missing `name:`?

website/docs/reference/opspec/op.yml/predicate.md
chrisdostert: currently yes

website/docs/reference/opspec/index.md
```diff
@@ -2,96 +2,29 @@
...
+
+An op is any directory containing a [valid opspec file with the name `op.yml`](op.yml/index). By default, the opctl cli will look for ops in the `.opspec/` directory in your current working directory, but ops can be defined in any directory.
+
+An op directory may contain other files and directories. These files are considered "embedded" within the op and can be referenced during execution as executable code, as data, or for whatever purpose you have.
+
+If the root of an op directory contains a `icon.svg` file (must be [SVG 1.1](https://www.w3.org/TR/SVG11) with a 1:1 aspect ratio), it will be used when displaying the op within a UI.
+
+## Call graph
+
+As an op runs, opctl parses [`op.yml`](op.yml/index) files to create a dynamic call graph tracking the different running containers as the are started and exit. The call graph is defined declaritively using [call objects](op.yml/call/index), which include sequencing and conditional support.
```
chrisdostert, line 16: Typo `as the are started`

website/docs/reference/opspec/index.md
```diff
@@ -2,96 +2,29 @@
...
+
+## Data flow
+
+Opctl has first class support for data flow into, between, and out of ops and their containers. Opspecs receive data from [`inputs`](op.yml#inputs) and return data through [`outputs`](op.yml#outputs). Within an opspec, data is referenced with a named variable that contains a typed value.
```
chrisdostert, line 20: Maybe consider `with a named variable that contains a typed value` => `via named variables containing typed values` (could also consider linking typed to the types section)

website/docs/reference/opspec/index.md
```diff
@@ -101,5 +34,55 @@ Reference:
 - [socket](types/socket.md)
 - [string](types/string.md)
 
-## Scoping
-Variables in opspec are scoped to each operation. [Parameters](../op/parameter/index.md) allow passing values into or out of this scope.
\ No newline at end of file
+### Scoping
+
+A variable is created when defined as an [`input`](op.yml#inputs) or [`output`](op.yml#inputs), bound implicitly to an `output` when making an [op call](op.yml/call/op), or declared with [loop variables](./op.yml/loop-variable). A variable name is in scope from creation through the lifetime of it's `op.yml` file. Variable scope does not extend into [op calls](./op.yml/call/op).
```
chrisdostert, line 39: This reads ambiguously to me `Variable scope does not extend into [op calls](./op.yml/call/op)`. Is it the current op call, child op call?

website/docs/reference/opspec/index.md
```diff
@@ -101,5 +34,55 @@ Reference:
...
+
+Variables can be referenced using the syntax `$(name)`.
+
+#### Scoping example
+
+```yml
```
chrisdostert, line 45: You might consider adding headings so these examples can be linked and so it's clear where one starts and the next begins

website/docs/reference/opspec/index.md
```diff
@@ -101,5 +34,55 @@ Reference:
...
+name: foobar
```
chrisdostert, line 46: You might consider making some or all of these trainings as opposed to reference.

website/docs/reference/opspec/op.yml/call/container/image.md
```diff
@@ -0,0 +1,37 @@
...
+
+## Properties
+
+### `ref`
+
+_required_
+
+A string referencing a local or remote image.
+
+#### OCI images
```
chrisdostert, line 15: This isn't clear to me if this is a property or what

website/docs/reference/opspec/op.yml/call/container/image.md
```diff
@@ -0,0 +1,37 @@
...
+
+A [variable reference](../../variable-reference.md) evaluating to a [directory](../../../types/dir) conforming to a [v1.0.1 OCI `image-layout`](https://github.com/opencontainers/image-spec/blob/v1.0.1/image-layout.md).
+
+```yaml
+ref: $(myOCIImageLayoutDir)
+```
+
+#### Docker images
```
chrisdostert, line 23: Same as above; This isn't clear to me if this is a property or what

chrisdostert:
@apexskier Thanks for the contribution; it looks like theres some nice changes in here. 

Is it possible to break this up into a few smaller iterations/PR's? This is a massive set of changes and I've spent over an hour and a half trying to look through everything and comment but it's too much change for me to meaningfully review and consider in one shot.

website/docs/reference/opspec/index.md
```diff
@@ -101,5 +34,55 @@ Reference:
...
+    - op:
+        ref: ../other-op
+        inputs:
+          input1: $(variable1)
+        outputs:
+          output1: $(variable2)
+          output2: $(variable4)
+```
+
+The variable `variable1`'s value is passed into `../other-op` where it will be stored in the variable `input1`. `../other-op` will not have access to the `variable1` or `variable2` names. After the op call completes the variable `variable2` will contain the value from `output1` that was produced by `../other-op`. Additionally, the variable `variable3` will be implicitly created from the output binding and will have the value from `output2` stored in it.
```
emilymcmahon, line 75:
```suggestion
The variable `variable1`'s value is passed into `../other-op` where it will be stored in the variable `input1`. `../other-op` will not have access to the `variable1` name. After the op call completes the variable `variable2` will contain the value from `output1` that was produced by `../other-op`. Additionally, the variable `variable4` will be implicitly created from the output binding and will have the value from `output2` stored in it.
```

emilymcmahon: After reading this section I still don't think I have a grasp of when/which variables are implicitly bound.

emilymcmahon (commented):
I think what you've done here are necessary for opctl as a product -- however, in my ideal world, these reference docs shouldn't be needed for 90% of the dev use cases. 

Thinking back to puppet days, what percentage of devs knew the internals of how puppet worked? The answer is pretty low. For the most common use cases you could pattern off examples. It's only when you were doing something unusual that you'd need to understand puppet modules, variable scopes, manifests, classes/subclasses, etc. 

What I personally would find the most valuable for me wrt opctl is a _central place_ for examples that covers the most common use cases.

chrisdostert:
> I think what you've done here are necessary for opctl as a product -- however, in my ideal world, these reference docs shouldn't be needed for 90% of the dev use cases.
> 
> Thinking back to puppet days, what percentage of devs knew the internals of how puppet worked? The answer is pretty low. For the most common use cases you could pattern off examples. It's only when you were doing something unusual that you'd need to understand puppet modules, variable scopes, manifests, classes/subclasses, etc.
> 
> What I personally would find the most valuable for me wrt opctl is a _central place_ for examples that covers the most common use cases.

Just wanted to chime in I agreed to what @emilymcmahon is getting at; Reference docs are *not* meant to be a learning tool but rather a lookup tool. Learning is mostly meant to happen in the training section of the docs [example](https://opctl.io/docs/training/containers/how-do-i-communicate-with-an-opctl-container). We need alot more trainings but I think the basis is sound. Personally I find both necessary and valuable.

website/docs/reference/opspec/index.md
apexskier: Good rewording. Given this flows directly into the types section I won't add a link

website/docs/reference/opspec/index.md
apexskier: The link here exists to make unambiguous. Literally, `op` calls in the `op.yml`. I'll reword though

website/docs/reference/opspec/index.md
apexskier: Line 43 is the heading - this is all one example

website/docs/reference/opspec/op.yml/call/container/image.md
apexskier: It's a sub-heading of the `ref` property - there are two types, these two headings describe the two types.

website/docs/reference/opspec/index.md
apexskier: Me neither! After my initial pass at this, I discovered that variables can be implicitly bound in loop vars.

This is a start at an explanation and will give someone a reference point to ask questions about and make improvements. I think this is a good place case for a more in depth "training" page.

apexskier:
I understand this is a large amount of content, but this actually _is_ my attempt as limiting scope. I intentionally have not touched the homepage, cli, training, and other documentation because I wanted to focus on the opspec syntax and how it's used. I agree that it's pretty large, but I also had the goal of making the opspec documentation feel cohesive, which is why I decided not to take it one piece at a time. (I also suspect splitting this up more is going to massively expand the time it takes to review and get improvements made)

I also intentionally have limited the examples here. While I definitely think the training documentation should be expanded with more complete examples and explanations, I wanted to include examples within the opspec documentation that provide an anchoring point for readers like me who look at this more technical specification relatively early on. (tl;dr - the examples here are explicitly more minimal than what I'd like to add to the training documentation)

Overall, I'd like to think this PR generally improves something that currently is hard to understand. It's not meant to be the final, best version of opctl's documentation, but an improvement.

Another large part of why I did this was to give myself a better understanding of opctl. I didn't know the details of several parts of the opspec going into it, but I think writing this as the person learning how it works allows me to provide more detail on the specific pieces that are more valuable. I also feel like I need to understand the opspec syntax and runtime behavior before I'd be able to effectively write longer-form "training" documents.

website/docs/reference/opspec/op.yml/call/container/image.md
chrisdostert: Yeah I understand but that's not how it displays if you look at the rendered markdown; that's why it used to be a list of references to the actual definitions cuz then you don't have this ambiguity. I kinda think I prefer the former setup linking to the definitions for this reason.

website/docs/reference/opspec/index.md
chrisdostert: I personally can't follow it if that's the case; it all flows together and I lose track of what's happening.

chrisdostert:
> I understand this is a large amount of content, but this actually _is_ my attempt as limiting scope. I intentionally have not touched the homepage, cli, training, and other documentation because I wanted to focus on the opspec syntax and how it's used. I agree that it's pretty large, but I also had the goal of making the opspec documentation feel cohesive, which is why I decided not to take it one piece at a time. (I also suspect splitting this up more is going to massively expand the time it takes to review and get improvements made)
> 
> I also intentionally have limited the examples here. While I definitely think the training documentation should be expanded with more complete examples and explanations, I wanted to include examples within the opspec documentation that provide an anchoring point for readers like me who look at this more technical specification relatively early on. (tl;dr - the examples here are explicitly more minimal than what I'd like to add to the training documentation)
> 
> Overall, I'd like to think this PR generally improves something that currently is hard to understand. It's not meant to be the final, best version of opctl's documentation, but an improvement.
> 
> Another large part of why I did this was to give myself a better understanding of opctl. I didn't know the details of several parts of the opspec going into it, but I think writing this as the person learning how it works allows me to provide more detail on the specific pieces that are more valuable. I also feel like I need to understand the opspec syntax and runtime behavior before I'd be able to effectively write longer-form "training" documents.

I appreciate all of that but the trouble for me. is, with all of it at once I personally can't actually tell what's an improvement and what might regress because so much changed. With smaller incremental changes it becomes viable to compare what we have to what's proposed and clearly understand the pros/cons of the proposed changes and provide useful feedback.

website/docs/reference/opspec/index.md
chrisdostert: Anywhere that you would write `blah : $(blah)`, you can write `blah:`. That's when/which : ). The whole point behind implicit vars was literally to save you having to repeat yourself; that's it.

website/docs/reference/opspec/op.yml/call/container/image.md
gunnarg-remitly: I've been reading through these and I think what I'd suggest is a list of properties under the "properties" heading that links to each sub-heading.  I think that would add clarity that "these are the properties, this subset is required, and here is a more in-depth description of each property"

website/docs/reference/opspec/op.yml/predicate.md
```diff
@@ -0,0 +1,145 @@
...
+
+```yml
```
gunnarg-remitly, line 11: I really like having examples for each documented thing.  I think it might make it slightly more accessible if the examples followed a standard format, where they are labelled as an example, the purpose is briefly described, the example is presented and the expected behavior is summarized.

It would add some length to the docs, but I think it'd add a lot of clarity to the examples

chrisdostert:
This has been open for over a year and has gone stale. I'm going to close the PR but leave the branch

PR #912, opened 2021-04-08
Support multiline text in `opctl ls`

apexskier:
Before:

```
REF	DESCRIPTION
fix	Attempts to fix source code linting issues.

lint	Lints source code, including formatting and in future static analysis.

Run the `fix` op to automatically fix what's fixable.

release	Releases forge by:  
1) compiling
2) uploading darwin/linux binaries to s3
3) creating a github release

test	runs unit tests
compile	turns source into runnable artifacts
```

After:

```
REF	DESCRIPTION
lint	Lints source code, including formatting and in future static analysis.
	
	Run the `fix` op to automatically fix what's fixable.
release	Releases forge by:  
	1) compiling
	2) uploading darwin/linux binaries to s3
	3) creating a github release
test	runs unit tests
compile	turns source into runnable artifacts
fix	Attempts to fix source code linting issues.
```

This could definitely be better - it's hard to tell where the description for one op ends and the next begins, but it's definitely an improvement.

### DIFF
--- a/cli/ls.go
+++ b/cli/ls.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"bufio"
 	"context"
 	"fmt"
 	"os"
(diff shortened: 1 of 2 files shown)


apexskier:
Something's up with code coverage - this should increase, since it adds tests for `ls.go`

chrisdostert (approved):
Nice iterative ehancement!

Re: future enhancements; some ideas i've had/heard have been:

1. Support rendering markdown
2. Consider UX; should it be similar to a manpage where you can scroll and have multiline?

apexskier:
1, sure. For 2, I'd prefer to leave that to another tool (pipe the output to cat/bat/etc)

PR #918, opened 2021-04-22
provide example usage

gunnarg-remitly:
This PR provides a working example op the provides and uses a string input, to illustrate how to use string inputs in practice


apexskier (approved):
I've got a large PR refactoring this documentation, if you're able to review - https://github.com/opctl/opctl/pull/910.

PR #921, opened 2021-04-23
Skip pulling tagged, non-latest, already pulled images

### LINKED ISSUE #920: Don't docker pull unless necessary
## 💥 Proposal

We currently experience a high number of docker rate limiting errors (INFRA-1870), which I suspect is due to opctl's policy of pulling every time a container is run. I've hit this enough that I've applied this change to my branch - https://github.com/opctl/opctl/commit/1fdde9f63ae7b96fa5debbc8bee974f289194a60.

I'd propose we skip pulling if (a) the image is tagged and (b) the image has been pulled locally.

apexskier:
The goal here is to help avoid rate limiting errors we've been seeing frequently in CI builds, which are causing significant friction. Given CI nodes are swapped out regularly and we'll still always pull tagged or latest tagged images, this should be pretty resilient.

Resolves #920, INFRA-1870

### DIFF
--- a/sdks/go/node/core/containerruntime/docker/imagePuller.go
+++ b/sdks/go/node/core/containerruntime/docker/imagePuller.go
@@ -3,8 +3,12 @@ package docker
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 	"io"
+	"strings"
+	"time"
 
+	"github.com/docker/distribution/reference"
 	"github.com/docker/docker/api/types"
 	dockerClientPkg "github.com/docker/docker/client"
 	"github.com/docker/docker/pkg/jsonmessage"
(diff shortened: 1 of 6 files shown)


sdks/go/node/core/containerruntime/docker/imagePuller.go
```diff
@@ -38,12 +40,41 @@ type _imagePuller struct {
...
+					Data:        []byte(fmt.Sprintf("Skipping image pull: %s\n", imageRef)),
+					OpRef:       containerCall.OpPath,
+					ContainerID: containerCall.ContainerID,
+					RootCallID:  rootCallID,
+				},
+			})
+
+			return nil
+		}
+	}
```
lmarszal, line 77: Suggestion: wrap this whole new code into function (`func checkImageExistsAndUpToDate(...)` or sth)

lmarszal (approved):
Nice improvement. One small readability comment.

PR #924, opened 2021-04-28
Add example of string parameter

gunnarg-remitly:
I added an example to a string parameter. Originally this PR, but since I haven't cloned the project it's easier for me to make a new pr in the right file than it is to fix the old pr.


chrisdostert (approved):
LGTM

Issue #925, opened 2021-04-29
"edit this page" link in documentation is broken


gunnarg-remitly:
## 🐛 Bug Report

The "edit this page" link leads to the deprecated "master" branch instead of "main"

![image](https://user-images.githubusercontent.com/45185818/116591382-98cbf000-a8d3-11eb-8f6a-080475be94bc.png)

Before reporting an issue please remember:
-  One issue, one bug: Please report a single bug per issue.

(Write your answer here.)

## To Reproduce

1. go here: https://opctl.io/docs/reference/opspec/op-directory/op/parameter/array
2. click "edit this page"
3. get a 404

## Expected behavior

I thought it would open a github editor for the markdown file backing the page in question.

## Actual Behavior

it took me to github's 404 page

gunnarg-remitly:
this pr ought to fix this: https://github.com/opctl/opctl/pull/929

PR #929, opened 2021-04-29
change edit url from `master` to `main`

gunnarg-remitly:
opctl no longer uses the master branch.  The edit links all point to the master branch.  They should instead point to main.

issue: https://github.com/opctl/opctl/issues/925


chrisdostert (approved):
LGTM

PR #931, opened 2021-04-29
Update the number parameter doc with an example

gunnarg-remitly:
Make an example op that echos a number parameter

### DIFF
--- a/website/docs/reference/opspec/op-directory/op/parameter/number.md
+++ b/website/docs/reference/opspec/op-directory/op/parameter/number.md
@@ -18,3 +18,20 @@ A number to use as the value of the parameter when no argument is provided.
 
 ### isSecret
 A boolean indicating if the value of the parameter is secret. This will cause it to be hidden in UI's for example. 
+
+## Example
+
+This op echos a number parameter, defaulting to 9001, when run.
+
+```yaml
+name: example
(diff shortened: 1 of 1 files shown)


website/docs/reference/opspec/op-directory/op/parameter/number.md
```diff
@@ -18,3 +18,20 @@ A number to use as the value of the parameter when no argument is provided.
...
 ### isSecret
 A boolean indicating if the value of the parameter is secret. This will cause it to be hidden in UI's for example. 
+
+## Example
+
+This op echos a number parameter, defaulting to 9001, when run.
+
+```yaml
+name: example
+description: an example op
```
chrisdostert, line 28: Just a passing thought; you could consider moving the description to the op description; I don't know if this would actually work well or not though.

chrisdostert (approved):
LGTM

monknomo:
I don't think that'd add a lot to the op in this context, however, the more
I write these, the more I think a repo full of these examples that could be
used with tutorials would be pretty handy.

chrisdostert:
@gunnarg-remitly can we either merge or close? This has been open awhile.

chrisdostert (approved):

PR #932, opened 2021-04-29
Provide example object input parameter

gunnarg-remitly:
This is an example of using an object as an input parameter.  The example echos values of the object's keys to demonstrate how to access a top level value, a sub-object value and an array value within an object.

### DIFF
--- a/website/docs/reference/opspec/op-directory/op/parameter/object.md
+++ b/website/docs/reference/opspec/op-directory/op/parameter/object.md
@@ -18,3 +18,33 @@ An object to use as the value of the parameter when no argument is provided.
 
 ### isSecret
 A boolean indicating if the value of the parameter is secret. This will cause it to be hidden in UI's for example. 
+
+## Example
+The op uses an object input, and echos the values of the object's keys.
+
+```yaml
+inputs:
+    example-input:
+        object:
+            default:
+                keyA: "valueA"
+                keyB: ["subValueB1", "subValueB2", "subValueB3"]
+                keyC:
+                    subKeyC1: "subValueC1"
+                    subKeyC2: "subKeyC2"
+run:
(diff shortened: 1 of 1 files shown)


website/docs/reference/opspec/op-directory/op/parameter/object.md
```diff
@@ -18,3 +18,33 @@ An object to use as the value of the parameter when no argument is provided.
...
+        object:
+            default:
+                keyA: "valueA"
+                keyB: ["subValueB1", "subValueB2", "subValueB3"]
+                keyC:
+                    subKeyC1: "subValueC1"
+                    subKeyC2: "subKeyC2"
+run:
+    container:
+        image: { ref: 'apteno/alpine-jq' }
```
chrisdostert, line 37: might consider using straight alpine

chrisdostert (approved):
LGTM

website/docs/reference/opspec/op-directory/op/parameter/object.md
gunnarg-remitly: yeah, I was initially thinking I would get valid json out of the object and that I could format it, but that is not the case

chrisdostert: objects/arrays do get serialized to JSON but if you're injecting it in a shell you'll need to make sure to escape it since quotes have special meaning to shells.

chrisdostert: @gunnarg-remitly  any update here?

chrisdostert:
@gunnarg-remitly can we either merge or close? This has been open awhile.

chrisdostert (approved):

PR #958, opened 2021-08-07
Allow using type initializers in input/output defaults; deprecate cur…

### LINKED ISSUE #957: Allow using type initializers in input/output defaults; deprecate current file/dir absolute path semantics
## 💥 Proposal

Currently if a file/dir typed parameter default is set to an absolute path. it is resolved from the directory containing the defining `op.yml`.

Aside from this, no other typed parameter defaults support type initialization, only verbatim values for the given type.

It's proposed that we deprecate the current file/dir default absolute path semantics, and instead allow using standard type initialization semantics. The scope for defaults would only include `./` and `../` since inputs would not exist in scope yet (we're interpreting them). 

The benefits would be:
- less to know; you wouldn't need to know a special syntax only supported in parameter defaults.
- allows all the benefits of type initializers see [array initialization](https://opctl.io/docs/reference/opspec/types/array#initialization), [boolean initialization](https://opctl.io/docs/reference/opspec/types/boolean), [dir initialization](https://opctl.io/docs/reference/opspec/types/dir#initialization), [file initialization](https://opctl.io/docs/reference/opspec/types/file#initialization), [number initialization](https://opctl.io/docs/reference/opspec/types/number#initialization), [object initialization](https://opctl.io/docs/reference/opspec/types/object#initialization), [string initialization](https://opctl.io/docs/reference/opspec/types/string#initialization) 

## Example
before
```yaml
inputs:
  myInput:
    string:
      default: /
```

after
```yaml
inputs:
  myInput:
    string:
      default: $(./)
```

chrisdostert:
…rent file/dir absolute path semantics; closes #957


chrisdostert:
FYI I was running into some issues asserting expected errors due to use of errors.Wrap so I replaced those with fmt.Errorf which is supported natively by gomega's `MatchError` assertion. I think we should remove all uses of `github.com/pkg/errors` in favor of standard golang errors but that can be a separate proposal/PR.

chrisdostert:
> FYI I was running into some issues asserting expected errors due to use of `github.com/pkg/errors.Wrap` so I replaced those with `fmt.Errorf` which is supported natively by gomega's `MatchError` assertion. I think we should remove all uses of `github.com/pkg/errors` in favor of standard golang errors but that can be a separate proposal/PR.

[here it is](https://github.com/opctl/opctl/pull/959)

tallen94 (approved):

PR #959, opened 2021-08-08
Remove github.com/pkg/errors usage; use native golang errors

chrisdostert:
Why?
- github.com/pkg/errors [is in maintenance mode](https://github.com/pkg/errors#roadmap)
- using native go errors functionality [is more idiomatic in golang](https://blog.golang.org/go1.13-errors#TOC_5.)
- removes a 3rd party dep

### DIFF
--- a/sdks/go/opspec/interpreter/call/op/interpret_test.go
+++ b/sdks/go/opspec/interpreter/call/op/interpret_test.go
@@ -1,0 +1,77 @@
+package op
+
+import (
+	"context"
+	"encoding/json"
+	"io/ioutil"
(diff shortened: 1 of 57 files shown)


sdks/go/opspec/interpreter/call/op/interpret_test.go
```diff
@@ -0,0 +1,77 @@
+package op
```
chrisdostert, line 1: unrelated; just added to meet codecoverage requirements

psamaan (approved):
lgtm

PR #961, opened 2021-08-08
Live progress & log line prefixes from opctl run

### LINKED ISSUE #664: CLI: No way to tell which op output comes from
Opctl's output is really difficult to understand right now, since there's no way to tell which op is emitting which output (either visually, or through `| grep` or something).

A simple way to improve this would be by prefixing each line with the op's name and ID.

A good example of this done well (for me), is the tool `concurrently`'s ["Prefix styling" options](https://github.com/kimmobrunfeldt/concurrently#usage).

Related to #62

chrisdostert:
When running an op via opctl run:
  - display progress via a live call graph
  - prefix log lines emitted by workloads with their op id & ref; closes #664

Note: this is basically a minimal port of the above features from @apexskier 's [branch](https://github.com/opctl/opctl/tree/mini-opctl). 

Just want to say a huge round of applause and thanks to @apexskier . This is some super cool & useful functionality.

### DIFF
--- a/cli/cli.go
+++ b/cli/cli.go
@@ -14,6 +14,7 @@ import (
 	"github.com/opctl/opctl/cli/internal/nodeprovider/local"
 	"github.com/opctl/opctl/sdks/go/model"
 	"github.com/opctl/opctl/sdks/go/opspec"
+	"golang.org/x/term"
 )
 
 var testModeEnvVar = "OPCTL_TEST_MODE"
@@ -278,6 +279,7 @@ func newCli(
 	cli.Command("run", "Start and wait on an op", func(runCmd *mow.Cmd) {
 		args := runCmd.StringsOpt("a", []string{}, "Explicitly pass args to op in format `-a NAME1=VALUE1 -a NAME2=VALUE2`")
 		argFile := runCmd.StringOpt("arg-file", filepath.Join(opspec.DotOpspecDirName, "args.yml"), "Read in a file of args in yml format")
+		noProgress := runCmd.BoolOpt("no-progress", !term.IsTerminal(int(os.Stdout.Fd())), "Disable live call graph for the op")
 		opRef := runCmd.StringArg("OP_REF", "", "Op reference (either `relative/path`, `/absolute/path`, `host/path/repo#tag`, or `host/path/repo#tag/path`)")
 
 		runCmd.Action = func() {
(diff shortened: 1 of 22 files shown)


cli/cli.go
```diff
@@ -278,6 +279,7 @@ func newCli(
 	cli.Command("run", "Start and wait on an op", func(runCmd *mow.Cmd) {
 		args := runCmd.StringsOpt("a", []string{}, "Explicitly pass args to op in format `-a NAME1=VALUE1 -a NAME2=VALUE2`")
 		argFile := runCmd.StringOpt("arg-file", filepath.Join(opspec.DotOpspecDirName, "args.yml"), "Read in a file of args in yml format")
+		noProgress := runCmd.BoolOpt("no-progress", !term.IsTerminal(int(os.Stdout.Fd())), "Disable live call graph for the op")
```
apexskier, line 282: I think it might be worth making this false by default while we test this on the main branch.

apexskier (commented):
Looks good. There's a known bug on my branch when a prompt for an input ("Empty call graph" is shown and sort of obscures the prompt) but that doesn't appear to be an issue here.

Tested with this op:
```
name: test
inputs:
  test:
    array:
      description: https://golang.org/cmd/go/#hdr-Build_and_test_caching
run:
  container:
    cmd: [echo, $(test)]
    image: { ref: "alpine" }
```

chrisdostert:
> Looks good. There's a known bug on my branch when a prompt for an input ("Empty call graph" is shown and sort of obscures the prompt) but that doesn't appear to be an issue here.
> 
> Tested with this op:
> 
> ```
> name: test
> inputs:
>   test:
>     array:
>       description: https://golang.org/cmd/go/#hdr-Build_and_test_caching
> run:
>   container:
>     cmd: [echo, $(test)]
>     image: { ref: "alpine" }
> ```

I touched up a couple of things like this as I was porting. There's a couple of outstanding ones I noticed that I haven't (yet) spent time looking into. I think this is already so useful I'd rather get it in and then we can file those as issues and address separately since they're non-critical cosmetic things. They were:

1. Certain log lines don't get prefixed (It's almost certainly the content of those log lines; seems to always occur in yarn installs for example which generally have fancy CLI UX).
2. Vertically resizing the terminal results in progress graph lines getting mixed in with the log lines.

apexskier (approved):

PR #972, opened 2021-08-19
Fix download links that are off

chrisdostert:


apexskier:
I beat you to it - https://github.com/opctl/opctl/pull/971

PR #981, opened 2021-10-13
Update website docs to note self-update won't work for versions older…

chrisdostert:
… than 0.1.48


chrisdostert:
> Why do you have to delete the data dir?

not necessarily required but just in case you upgrade from a version that had older event schema; it's a sure way to not run into a scenario you have unsupported historical state.

PR #983, opened 2021-11-03
Stop Docker containers gracefully before removing them

ehsanul:
This notifies containers with a SIGTERM signal, giving them a chance to gracefully shut down within a 3s timeout window. Containers that fail to stop within that window will be forced to terminate with a SIGKILL.

Note that the k8s runtime's pod deletion defaults to a graceful termination, so this change also brings the Docker runtime into alignment.

### DIFF
--- a/sdks/go/node/core/containerruntime/docker/runContainer.go
+++ b/sdks/go/node/core/containerruntime/docker/runContainer.go
@@ -2,12 +2,13 @@ package docker
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"io"
 	"strings"
 	"sync"
+	"time"
 
-	"errors"
 	"github.com/docker/docker/api/types"
 	"github.com/docker/docker/api/types/container"
 	"github.com/docker/docker/api/types/network"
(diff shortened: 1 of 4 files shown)


sdks/go/node/core/containerruntime/docker/runContainer.go
```diff
@@ -82,9 +83,16 @@ func (cr _runContainer) RunContainer(
 	// do not change this prefix as it might break external consumers
 	containerName := getContainerName(req.ContainerID)
 	defer func() {
-		// ensure container always cleaned up
+		// ensure container always cleaned up: gracefully stop then delete it
+		newCtx := context.Background() // always use a fresh context, to clean up after cancellation
+		stopTimeout := 3 * time.Second
+		cr.dockerClient.ContainerStop(
```
chrisdostert, line 89: Maybe consider calling DeleteContainerIfExists here to consolidate the logic to a single place

ehsanul: I gave that a shot, but found that we don't have the `_containerRuntime` in this context, just the `_runContainer`, which is a field of `_containerRuntime`. I don't really understand the split between these two structs: the docker client is present on both the parent and child structs and the parent struct has nothing else in it, so they seem redundant. The `RunContainer` method is present on both the `runContainer` interface and the `ContainerRuntime` interface and I don't understand how these relate to each other, if at all.

I'm not sure of the best way to refactor to consolidate the logic. My hunch is that the docker runtime needs to implement `RunContainer` directly on `_containerRuntime` and `_runContainer` needs to go away, but this sounds like a large refactor. Do you consider this blocking for this PR?

chrisdostert: They're just split for testability so we don't end up with a godmother struct. It's not blocking but just was an optimization. We can revisit at a later time. You might just mention in an @TODO comment they should be consolidated.

chrisdostert (approved):

ehsanul:
@Remitly/bot merge

sdks/go/node/core/containerruntime/docker/runContainer.go
ehsanul: Added the @TODO comment, please go ahead and merge when you have a chance. Thanks!

PR #987, opened 2021-12-05
VM management

chrisdostert:
# What's in this PR
TLDR; this PR adds support for a new container runtime, namely 'qemu', which among other benefits removes any need to use docker for mac. 

Additionally it:
1. introduces a new command `opctl node delete` which deletes a node and any associated state
2. ~~renames `opctl node kill` => `opctl node stop` and `opctl node create` => `opctl node start` as the terms `create` and `kill` have always seemed ambiguous~~

## Unrelated to feature
1. moved `github.com/opctl/opctl/sdks/go/node/core` to `github.com/opctl/opctl/sdks/go/node`
2. fixed go package cache mounted to wrong directory in ops
3. fixed invalid state if cloning op interrupted midway
4. removed unnecessary go routine wrapping and arbitrary channel buffering from eventStore

# State
There are some [known limitations](https://github.com/opctl/opctl/issues/988).


tallen94:
- Is there default behavior for Mac? 
- Are we still allowed to use docker for the container runtime?
- Does this only work for specific versions of OSX?

Also, I might consider making the command naming changes separate, they also seem unnecessary from a user perspective unless you have seen people get confused. Like I am comfortable with the current command names (unless its required to enable the qemu changes)

chrisdostert:
> * Is there default behavior for Mac?

The default is still docker on Mac OS because of the linked limitations

> * Are we still allowed to use docker for the container runtime?

Yep, they work side by side; not mutually exclusive.

> * Does this only work for specific versions of OSX?
It should work for all recent versions of OSX

sdks/go/node/api/handler/events/stream/handler.go
```diff
@@ -59,7 +59,7 @@ func (hdlr _handler) Handle(
 	if sinceString := httpReq.URL.Query().Get("since"); sinceString != "" {
 		sinceTime, err := time.Parse(time.RFC3339, sinceString)
 		if err != nil {
-			http.Error(httpResp, err.Error(), http.StatusBadRequest)
```
tallen94, line 62: Are these errors going to get lost?

sdks/go/node/containerruntime/docker/imagePuller.go
```diff
@@ -66,7 +66,9 @@ func (ip _imagePuller) Pull(
 	imagePullCreds := containerCall.Image.PullCreds
 	containerID := containerCall.ContainerID
 
-	imagePullOptions := types.ImagePullOptions{}
+	imagePullOptions := types.ImagePullOptions{
+		Platform: "linux",
```
tallen94, line 70: Why is this assumed now?

sdks/go/node/api/handler/events/stream/handler.go
chrisdostert: They never were used; the connection is upgraded to a websocket above so these all were doing nothing other than logging that we're (invalidly) attempting to writing to hijacked connection .

sdks/go/node/containerruntime/docker/imagePuller.go
chrisdostert: I noticed it was being mentioned in logs it was looking for an image with Platform "darwin" and falling back to "linux" so It seemed "more correct" to specify linux. I can't think of a case we wouldn't want it to be that.

tallen94 (approved):

PR #995, opened 2021-12-18
Make node locking mechanism more reliable; closes #913

### LINKED ISSUE #913: Node locking mechanism doesn't ensure process is opctl
## 🐛 Bug Report

When I want to run opctl command I receive the following message:

```
$ opctl run dev
Post "http://127.0.0.1:42224/api/ops/starts": dial tcp 127.0.0.1:42224: connect: connection refused
````

I tried to run this command but without luck

```
$ opctl events
dial tcp 127.0.0.1:42224: connect: connection refused
```

opctl Version: 0.1.47
docker up and running (restarted)
I'm running on MacOS

### PR DESCRIPTION
**Pull request by chrisdostert**
This PR moves the daemons underlying file lock to using built in OS file locks. This solves for an edge case where the opctl daemon may no longer be running but the PID in it's pid.lock happens to be held by a running process. By using built in OS file locks, we can ensure the lock is removed when the opctl daemon process exits. 

closes #913

### CODE REVIEW

File: sdks/go/node/api/client/client.go [hunk 1]
```diff
@@ -25,22 +25,9 @@ func New(
 	baseURL url.URL,
 	opts *Opts,
 ) node.Node {
-
```
zachpuck, line 28: Just curious: way this change?

zachpuck (approved):

sdks/go/node/api/client/client.go
```diff
@@ -25,22 +25,9 @@ func New(
...
-	httpClient.MaxRetries = 90
-
-	if opts != nil {
-		// handle options
-		httpClient.LogHook = func(errEntry pester.ErrEntry) {
-			// wire up retry log hook
-			opts.RetryLogHook(errEntry.Err)
-		}
-	}
-
```
zachpuck, line 40: Just curious: way this change?

cli/internal/nodeprovider/local/createNodeIfNotExists.go
```diff
@@ -76,9 +69,38 @@ func (np nodeProvider) CreateNodeIfNotExists(ctx context.Context) (node.Node, er
...
+	go func() {
+		for {
+			select {
+			case <-ctx.Done():
+				return
+			default:
+				if err := apiClientNode.Liveness(ctx); err == nil {
+					cancel()
+				}
+				time.Sleep(time.Second)
```
chrisdostert, line 83: @zachpuck replaced with this loop. This was the only place we "relied" on that retry behavior from that 3p lib, so I thought reasonable to simplify. It also helps startup faster because the 3p lib does exponential backoff whereas here we just do linear.

PR #1035, opened 2022-10-06
Fix crash on bad slice input parameter

apexskier:
There's a crash that can be triggered when using an array input and forgetting to use the `name=` syntax on the CLI.

Before:

auth-testing — -bash — bash - Terminal - 2022-10-06 at 12 24 15@2x

After:

auth-testing — -bash — bash - Terminal - 2022-10-06 at 12 24 20@2x

### DIFF
--- a/cli/internal/cliparamsatisfier/inputsrc/slice/slice.go
+++ b/cli/internal/cliparamsatisfier/inputsrc/slice/slice.go
@@ -14,6 +14,9 @@ func New(
 		// get parts
 		parts := strings.SplitN(arg, sep, 2)
 		inputName := parts[0]
+		if len(parts) < 2 {
+			continue
+		}
 		inputValue := parts[1]
 
 		argMap[inputName] = inputValue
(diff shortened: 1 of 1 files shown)


cli/internal/cliparamsatisfier/inputsrc/slice/slice.go
```diff
@@ -14,6 +14,9 @@ func New(
 		// get parts
 		parts := strings.SplitN(arg, sep, 2)
 		inputName := parts[0]
+		if len(parts) < 2 {
```
chrisdostert, line 17: can `parts[0]` npe? If so should we move this check up a line?

chrisdostert (approved):

cli/internal/cliparamsatisfier/inputsrc/slice/slice.go
apexskier: [Technically yes.](https://go.dev/play/p/KT4esa5e8kp) I don't think it can in real usage, since I don't think it's possible to get it to be an empty string with the way arg parsing works.

But, it definitely makes sense to move this, it won't have a negative impact.

PR #1057, opened 2023-04-30
Automatic GPU passthrough, Logging enhancements, migrate e2e test harness from sharness to opctl

chrisdostert:
This PR:
- updates from golang 1.18 to 1.20
- adds) Automatic GPU detection/passthrough for docker runtime
- fixes) `lte` predicate
- fixes) `opctl ls` renders ops with no description as blank line
- fixes) In some cases, stacktraces logged as bytes rather than strings
- fixes) Containers aren't automatically removed after exit for docker runtime
- fixes) In some cases, `Runtime error: invalid memory address or nil pointer dereference` error occurs
- fixes) CLI not logging errors from the root op
- replaces sharness with opctl as e2e test runner
  - cons:
    - runs slower since it spins up a container with an embedded docker daemon and opctl node for every scenario. Could run in parallel but will keep serial since debugging failures is simpler
  - pros:
    - another way to dogfood
    - better isolation between tests
- adds tests to the test-suite for `lt`, `lte`, `gt`, `gte` predicates
- adds training for flow i.e. conditional and looped execution


cli/cli.go
```diff
@@ -40,7 +40,7 @@ func newCli(
 		} else {
 			cliOutput.Error(err.Error())
 
-			if re, ok := err.(*RunError); ok {
+			if re, ok := err.(RunError); ok {
```
zachpuck, line 43: Just Curious was this the nil pointer bug we where seeing sometimes?

zachpuck (approved):

cli/internal/opgraph/callgraph.go
```diff
@@ -37,6 +37,9 @@ var errNotFoundInGraph = errors.New("not found in graph")
 const skippedState = "skipped"
 
 func (n *callGraphNode) insert(call *model.Call, startTime time.Time, initialState string) error {
+	if n == nil || n.call == nil || call == nil {
```
chrisdostert, line 40: @zachpuck this was the nil pointer bug we saw sometimes; it's got to do with event order not being guaranteed so n can be nil.

cli/cli.go
chrisdostert: Not here I just mentioned you where that was. The reason I removed the pointer from here was I just didn't see any reason for this being a pointer and leaving a potential for a NPE.

PR #1098, opened 2024-04-19
Check whether Changelog is included in diff, lint changelog on main and in PR

chris-spalding:
# Overview
It would be cool to move opctl to CD. As a first step I'd like to introduce the following changes to CI/CD:
1. Add a job in CI that runs on PRs only. If the changelog is included in the diff, pass. If the changelog is not included then fail (but don't block CI) and print a warning message along the lines of ("WARNING: a new release will not be created when these changes are merged to main unless you add a new entry to CHANGELOG.md")
2. Add a job in CI that lints the current CHANGELOG to make sure it matches the keep-a-changelog format.

In a subsequent PR I'd like to add the following changes:
1. Add a job to the main branch's workflow that runs when CHANGELOG.md was edited. Parse the first version encountered in the file and run the release op using that as the new version. The release will be created as a prerelease and will not be marked as latest, so a maintainer will need to manually update the release to mark it as latest and remove the prerelease tag when the change is good to go

## Gotchas
1. We need to sort out what docker creds to add to this repo's secrets so that the new op I'm adding can talk to dockerhub. Maybe we can make a service account on dockerhub for the opctl organization?
2. There are two potential issues on Codacy that were introduced by this PR, and I don't think they're actually issues based on the input that will be provided to the code that uses the regexes. You can see the issues here https://app.codacy.com/gh/opctl/opctl/pull-requests/1098/issues

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -7,6 +7,29 @@ on:
     branches: [ main ]
 
 jobs:
+  check-for-changelog:
+    name: Check for CHANGELOG.md
+    runs-on: ubuntu-latest
+    steps:
+    - name: Check out code
+      uses: actions/checkout@v4
+      with:
+        fetch-depth: 0
+    - name: Install Opctl
+      run: curl -L https://github.com/opctl/opctl/releases/latest/download/opctl-linux-amd64.tgz | sudo tar -xzv -C /usr/local/bin
+    - name: Check for CHANGELOG.md
+      run: opctl run check-for-changelog
+
+  lint-changelog:
+    name: lint-changelog
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v2
+      - name: Install Opctl
+        run: curl -L https://github.com/opctl/opctl/releases/latest/download/opctl-linux-amd64.tgz | sudo tar -xzv -C /usr/local/bin
+      - name: Run Markdownlint
+        run: opctl run lint
 
   build:
     name: Build
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -8,7 +8,8 @@ on:
 
 jobs:
   check-for-changelog:
-    name: Check for CHANGELOG.md
+    name: Check for CHANGELOG.md (PR)
+    if: github.event_name == 'pull_request'
     runs-on: ubuntu-latest
     steps:
     - name: Check out code
--- a/.markdownlint/.markdownlint.json
+++ b/.markdownlint/.markdownlint.json
@@ -1,0 +1,10 @@
+{
+  "default": true,
+  "MD024": { "siblings_only": true },
+  "MD013": {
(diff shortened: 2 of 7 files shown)


chris-spalding:
There were two regex issues that Codacy called out that I think can be safely ignored. I bypassed the checks but you can see the errors here if you'd like to: https://app.codacy.com/gh/opctl/opctl/pull-requests/1098/issues

chrisdostert:
This seems nice. Once thought regarding the changelog validator. Maybe we could use someone elses validator as opposed to maintaining our own. I quickly noticed [this one has a docker image](https://github.com/cyberark/parse-a-changelog/blob/e8b717408dae50350cd1646b6eb34e9fd1558af0/README.md?plain=1#L33) so would be pretty simple to use in an op; there could be even better though.

chris-spalding:
> This seems nice. One thought regarding the changelog validator. Maybe we could use someone elses validator as opposed to maintaining our own. I quickly noticed [this one has a docker image](https://github.com/cyberark/parse-a-changelog/blob/e8b717408dae50350cd1646b6eb34e9fd1558af0/README.md?plain=1#L33) so would be pretty simple to use in an op; there could be even better though.

Ah yeah, I meant to add a note about the choice around the validator in the PR. If you feel really strongly about it then I'm fine with using `parse_a_changelog`. I will say that after trying to use that tool, though, the grammar is extremely strict to the point of not being useful. As an example of the errors I ran into when I used `parse_a_changelog` locally:
1. The first line content _must_ be `# Changelog`. So `# Change Log` failed the check. Not a big deal but also seems like too tight of a requirement
2. the second line content _must_ be `All notable changes to this project will be documented in this file.`. That content was on the third line of the file and the error message was a bit confusing
3. line 4 _must_ contain `://keepachangelog.com/en/1.0.0/)` but the most recent version of `keepachangelog` is 1.1.0. This the first actual bad behavior
4. the tool doesn't support code blocks in changesets so [this block](https://github.com/opctl/opctl/blob/main/CHANGELOG.md#changed-12) would need to be changed
5. etc etc

The main reason that I like markdownlint is that it's quite flexible. The custom parsing rule I defined in this PR in JavaScript (I think) is pretty easy to understand and it allows us to implement the important parts of the keepachangelog grammar without having to match everything (which could also be viewed as a downside). Basically it comes down to whether you want us to adhere to the standard to the point that even extraneous newline characters will cause changelog linting to fail, or if you just want to make sure that the big bits of the format are followed (e.g. semantic versions defined in second headers, third headers have the expected values)

scottb-remitly (approved):
I'm excited about this work, thanks @chris-spalding

chrisdostert:
Supporting the spec rigidly doesn't bother me and I even probably prefer it, but the out of date reference isn't great. It's interesting how few validators seem to be out there. I'm good with rolling this and we can always iterate later by swapping to a 3p validator if one ever emerges.

chrisdostert (approved):

scottb-remitly (approved):
Good catch

PR #1102, opened 2024-05-30
Fix git repo collisions

keithjstone:
Demonstrate conflicting dataref git checkout and fix by correctly destroying whole git repo before clone.


chrisdostert:
resolved by #1124

PR #1103, opened 2024-06-05
do not set default version in ops

chris-spalding:
# Overview
Prior to this change, when someone runs the `compile` op in the CLI directory, their binary will be built with a version `0.0.1`. That's not desirable because new maintainers can very easily miss that they're expected to provide a new version when they compile the opctl CLI 😅. This change:
1. removes all default values for `version` inputs in ops
2. removes `version` as an input from the test op because that appears to have been introduced as a copy pasta (it's not used in the op at all)
3. Updates the `github.release.create` op to version 2.0.0 from 1.2.0

## Testing
1. I expect CI to continue to pass
2. Locally I ran the `compile` op in the CLI directory and verified that the built binary has a version that I expect.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -55,4 +55,4 @@ jobs:
     - name: Install Opctl
       run: curl -L https://github.com/opctl/opctl/releases/latest/download/opctl-linux-amd64.tgz | sudo tar -xzv -C /usr/local/bin
 
-    - run: opctl run -a gitBranch=${{ steps.branch_name_push.outputs.branch }}${{ steps.branch_name_pr.outputs.branch }} build
+    - run: opctl run -a gitBranch=${{ steps.branch_name_push.outputs.branch }}${{ steps.branch_name_pr.outputs.branch }} -a version=0.0.1 build
--- a/.opspec/build/op.yml
+++ b/.opspec/build/op.yml
@@ -11,7 +11,6 @@ inputs:
       default: main
   version:
     string:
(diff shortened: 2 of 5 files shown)


.github/workflows/build.yml
```diff
@@ -55,4 +55,4 @@ jobs:
     - name: Install Opctl
       run: curl -L https://github.com/opctl/opctl/releases/latest/download/opctl-linux-amd64.tgz | sudo tar -xzv -C /usr/local/bin
 
-    - run: opctl run -a gitBranch=${{ steps.branch_name_push.outputs.branch }}${{ steps.branch_name_pr.outputs.branch }} build
+    - run: opctl run -a gitBranch=${{ steps.branch_name_push.outputs.branch }}${{ steps.branch_name_pr.outputs.branch }} -a version=0.0.1 build
```
scottb-remitly, line 58: what's the purpose of adding `-a version=0.0.1` to the build command via this build action? It confuses me since the `build op` had the `version` property removed.

chris-spalding: The `test` op had `version` remove, not the build op. The build op had the default value for `version` removed. The only place that the `version` actually matters is during the release op - but it has to be defined in CI even though the release op isn't run in CI

scottb-remitly: I saw it removed from the `test/inputs` section and missed it still present in the op's `inputs` section. Collapsing PR reviews 😓

scottb-remitly (approved):

PR #1104, opened 2024-06-14
reorganize changelog ops and move from docker to ghcr

chris-spalding:
# Overview
1. Organize all ops related to the changelog under a single directory
2. Migrate from docker to ghcr, which simplifies opctl operationally since we don't have to deal with multiple distribution channels

## Testing
CI passes, and I was able to push a container to GHCR locally with the 0.1.54-pre tag


.opspec/release/op.yml
```diff
@@ -40,22 +40,25 @@ inputs:
       description: version of opctl being released
 run:
   serial:
+    - op:
+        ref: $(../changelog/get-latest-version)
+        outputs:
+          latestVersion:
     - op:
```
chrisdostert, line 47: do you think we need to add a condition here to skip if latestVersion == currentVersion ?

chrisdostert: right now it will run every time (and re-publish over existing) right?

chris-spalding: good call - I've updated the release op to first make sure a tag doesn't exist in GHCR

chris-spalding: Though wouldn't the same problem have existed in the previous implementation? I don't think the docker registry is immutable

chrisdostert: previously releases were manual so it was up to the releaser to ensure.

.opspec/release/to-ghcr/check.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/sh
+
+set -e
+
+apk add --update docker
+
+# only continue with release if tag doesn't exist
+if docker pull ghcr.io/opctl/opctl:${version}-dind; then
+  echo "Opctl Image for version '${version}' already exists"
+  exit 1
```
chrisdostert, line 10: I don't think you want to exit the container non-zero cuz that will fail the operation. I think you will want to use opctl native flow logic/conditions to just skip the rest of the release in the case the release already exists.

chrisdostert: [docs](https://opctl.io/docs/training/flow/conditional-execution) in case that helps

chrisdostert: in our case I feel like we'd add a new serial block and put the condition on that and then add a container or op that lists releases from github and if the current one already exists we skip (or something like that). Maybe we add a github.releases.list op, something similar to what we do here for [repos but for releases](https://github.com/opspec-pkgs/_.op-repos.standardize/blob/6713437f2b566ba41abf43af659a5a82b5ec6c10/op.yml#L44)

chrisdostert: I guess another idea would just be to export a variable from the container you added something like `/isUnreleased.json: $(isUnreleased)` and condition on that instead of exiting non-zero

chris-spalding: done!

.opspec/release/to-ghcr/op.yml
```diff
@@ -0,0 +1,95 @@
...
+    string:
+      constraints: { format: semver }
+      description: version being published
+  dockerSocket:
+    socket:
+      description: docker socket
+    default: /var/run/docker.sock
+outputs:
+  alreadyPublished:
+    number:
```
chrisdostert, line 27: nit) opinionated but boolean would be more idiomatic IMHO (& slightly simpler e.g. no need to specify constraints)

.opspec/release/to-ghcr/op.yml
```diff
@@ -0,0 +1,95 @@
...
+        maximum: 1
+run:
+  serial:
+    - op:
+        ref: github.com/opspec-pkgs/base64.encode#1.1.0
+        inputs:
+          rawValue: $(github.username):$(github.accessToken)
+        outputs:
+          encodedValue: $(b64GithubAuth)
+    - container:
```
chrisdostert, line 37: don't you want to do this inside release so you can skip both build, release to github and release to ghcr?

.opspec/release/to-ghcr/check.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/sh
+
+set -e
+
+apk add --update docker
+
+# only continue with release if tag doesn't exist
+if docker pull ghcr.io/opctl/opctl:${version}-dind; then
```
chrisdostert, line 8: nit) It could be nice to check for release without taking a dep on docker since we don't specifically rely on docker (curling GitHub releases API for example) but not a blocker.

.opspec/release/to-ghcr/op.yml
chris-spalding: Yep, good call. Moved it up.

.opspec/release/to-ghcr/op.yml
```diff
@@ -0,0 +1,57 @@
...
+      description: whether this version is already published
+run:
+  parallelLoop:
+    range:
+      - dind
+      - dood
+    vars:
+      value: $(imageVariant)
+    run:
+      if:
```
chrisdostert, line 26: just FWIW, you can put this on any call so it could be on parallelLoop as well

.opspec/build/op.yml
```diff
@@ -13,10 +13,15 @@ inputs:
     string:
       constraints:
         format: semver
+  alreadyPublished:
```
chrisdostert, line 16: you might consider conditionally calling this or not within the release op as opposed to making it a parameter since if you're calling build you generally would want to build or you just wouldn't call it.

chrisdostert: here's an idea of what I meant

```yaml
description: performs a release
name: release
inputs:
  dockerSocket:
    socket:
      description: docker socket to use as container runtime 
  gitBranch:
    string:
      constraints: { minLength: 1 }
      description: git branch the coverage should be associated with
      default: main
  github:
    object:
      constraints:
        properties:
          accessToken:
            minLength: 1
            type: string
            writeOnly: true
          username:
            minLength: 1
            type: string
        required: [accessToken, username]
      description: configuration required to interact w/ github
  HOME:
    dir:
      description: Home directory of caller; used to access go modules
  version:
    string:
      constraints: { format: semver }
      description: version of opctl being released
run:
  serial:
    - op:
        ref: $(../changelog/get-latest-version)
        outputs:
          latestVersion:
    - op:
        ref: github.com/opspec-pkgs/base64.encode#1.1.0
        inputs:
          rawValue: $(github.username):$(github.accessToken)
        outputs:
          encodedValue: $(b64GithubAuth)
    - container:
        image: { ref: alpine:3.20 }
        envVars:
          version: $(version)
        cmd:
          - sh
          - -c
          - /check.sh
        sockets:
          /var/run/docker.sock: $(dockerSocket)
        files:
          /alreadyPublished: $(alreadyPublished)
          /check.sh:
          /root/.docker/config.json:
            auths:
              ghcr.io:
                auth: $(b64GithubAuth)
    - if:
        - eq: [$(alreadyPublished), false]
      serial:
        - op:
            ref: $(../build)
            inputs:
              dockerSocket:
              gitBranch:
              HOME:
              version: $(latestVersion)
              alreadyPublished:
        - parallel:
            - op:
                ref: $(./to-ghcr)
                inputs:
                  version: $(latestVersion)
                  alreadyPublished: $(alreadyPublished)
                  encodedGithubAuth: $(b64GithubAuth)
                  dockerSocket:
            - op:
                ref: $(./to-github)
                inputs:
                  github:
                  version: $(latestVersion)
                  alreadyPublished: $(alreadyPublished)
```

.opspec/release/to-ghcr/check.sh
chrisdostert: yeah only a nit; like you said we could revisit later if we want

.opspec/build/op.yml
chris-spalding: done! Didn't realize you could do that

chrisdostert (approved):

.opspec/release/check.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/sh
+
+set -e
+
+apk add --update docker
+
+# only continue with release if tag doesn't exist
+if docker pull ghcr.io/opctl/opctl:${version}-dind; then
+  echo "Opctl Image for version '${version}' already exists"
+  echo true | tr -d '\n' > /alreadyPublished
```
chrisdostert, line 10:
```suggestion
  echo true > /alreadyPublished
```

.opspec/release/to-ghcr/op.yml
```diff
@@ -0,0 +1,51 @@
+name: to-docker
```
chrisdostert, line 1:
```suggestion
name: to-ghcr
```

.opspec/release/to-ghcr/op.yml
```diff
@@ -0,0 +1,51 @@
...
+description: releases opctl to [docker hub](https://hub.docker.com/r/opctl/opctl/)
```
chrisdostert, line 2:
```suggestion
description: releases opctl to [Github Container Registry](https://ghcr.io/opctl/opctl/)
```

.opspec/release/to-ghcr/op.yml
```diff
@@ -0,0 +1,51 @@
...
+inputs:
+  encodedGithubAuth:
+    string:
+      description: github auth encoded as base64
+  version:
+    string:
+      constraints: { format: semver }
+      description: version being published
+  dockerSocket:
```
chrisdostert, line 11: not needed/used correct?

chrisdostert, line 11:
```suggestion
```

.opspec/release/to-ghcr/op.yml
```diff
@@ -0,0 +1,51 @@
...
+    socket:
```
chrisdostert, line 12:
```suggestion
```

.opspec/release/to-ghcr/op.yml
```diff
@@ -0,0 +1,51 @@
...
+      description: docker socket
```
chrisdostert, line 13:
```suggestion
```

.opspec/release/to-ghcr/op.yml
```diff
@@ -0,0 +1,51 @@
...
+    default: /var/run/docker.sock
```
chrisdostert, line 14:
```suggestion
```

.opspec/release/op.yml
```diff
@@ -38,24 +29,58 @@ inputs:
...
+            gitBranch:
+            HOME:
+            version: $(latestVersion)
+      - parallel:
+          - op:
+              ref: $(./to-ghcr)
+              inputs:
+                version: $(latestVersion)
+                encodedGithubAuth: $(b64GithubAuth)
+                dockerSocket:
```
chrisdostert, line 81:
```suggestion
```

.opspec/changelog/get-latest-version/op.yml
```diff
@@ -0,0 +1,26 @@
...
+      - -c
+      - |
+        echo "getting latest version from Changelog"
+        sed -rn 's/^## ([0-9]+.[0-9]+.[0-9]+-?[A-Za-z0-9]{0,}) - .*$/\1/p' CHANGELOG.md | head -n 1 | tr -d '\n' > /latestVersion
+    workDir: /src
+    dirs:
+      /src: $(../../..)
+    files:
+      /latestVersion: $(latestVersion)
+      /root/.gitconfig: |
```
chrisdostert, line 24:
```suggestion
```

.opspec/changelog/get-latest-version/op.yml
```diff
@@ -0,0 +1,26 @@
...
+        [safe]
```
chrisdostert, line 25:
```suggestion
```

.opspec/changelog/get-latest-version/op.yml
```diff
@@ -0,0 +1,26 @@
...
+          directory = /src
```
chrisdostert, line 26:
```suggestion
```

.opspec/release/to-ghcr/op.yml
chris-spalding: presumably the last 3 comments were all supposed to be single suggestion - I got rid of the unused param

.opspec/changelog/get-latest-version/op.yml
chris-spalding: done

chrisdostert (dismissed):

.opspec/release/op.yml
```diff
@@ -38,24 +29,57 @@ inputs:
     string:
       constraints: { format: semver }
       description: version of opctl being released
+outputs:
```
chrisdostert, line 32:
```suggestion
```

.opspec/release/op.yml
```diff
@@ -38,24 +29,57 @@ inputs:
...
+  alreadyPublished:
```
chrisdostert, line 33:
```suggestion
```

.opspec/release/op.yml
```diff
@@ -38,24 +29,57 @@ inputs:
...
+    boolean:
```
chrisdostert, line 34:
```suggestion
```

.opspec/release/op.yml
```diff
@@ -38,24 +29,57 @@ inputs:
...
+      description: whether this version is already published
```
chrisdostert, line 35:
```suggestion
```

chrisdostert (approved):

Issue #976, opened 2021-09-20
Variable driven port binding


chrisdostert:
## 💥 Proposal
The current port binding syntax provides a way to expose an opctl container port to the host it's running on. 

A limitation of the current approach is that it breaks encapsulation; only one process at a time can bind a given host port so if a process is already running on the host on a port bound by a container in an op, it won't be able to start.

A major intention behind opctl is to provide a virtualized runtime; node processes should not effect the behavior of running an op.

The proposed enhancement to the existing syntax is to allow binding [socket typed](https://opctl.io/docs/reference/opspec/types/socket) variables to container ports (similar to how a container dir or file can be bound to variables). 

This would allow the caller of the op to choose where/if to mount the socket.

Example:
```yaml
name: improvedPorts
inputs:
  apiSocket:
    socket:
      default: localhost:80 # or could be name-to-add-to-hostfile:80 or /path/to/mount/as/unix/socket
run:
  container:
    image:
      ref: myApi
    sockets:
      80: $(apiSocket)
```

chrisdostert:
related https://github.com/opctl/opctl/issues/690

PR #1112, opened 2024-10-21
Access containers by their name from opctl nodes

chrisdostert:
This PR primarily adds ability to access containers by their name from opctl nodes.

It additionally adds direct container connectivity from opctl nodes running on Mac OS (already the default on all other OSs) which is required for the former on Mac OS.

It achieves this by:
- registering opctl as DNS server/resolver on the opctl node
- ensuring connectivity from opctl node to opctl managed containers (required to support docker for mac)

unrelated: 
- upgrades to golang 1.23
- moves `datadir`, `pubsub`,  packages within `node` package for consistency with newly introduced `dns` package location

relates to #976


.markdownlint/customRule.js
```diff
@@ -15,7 +15,7 @@ module.exports = {
           "lineNumber": heading.lineNumber,
           "detail": "First heading should be '# Change Log'.",
         });
-      } else if (heading.tag === "h2" && !/^## \d\.\d{1,9}\.\d{1,3}-?[A-Za-z0-9]{0,} - \d{4}-\d{2}-\d{2}$/.test(heading.line)) {
+      } else if (heading.tag === "h2" && !/^## \d\.\d{1,9}\.\d{1,3}(-[.A-Za-z0-9]+)? - \d{4}-\d{2}-\d{2}$/.test(heading.line)) {
```
chrisdostert, line 18: Unrelated but fixing to more fully support semver pre-release pattern (so we can use it in this PR)

cli/internal/nodeprovider/local/createNodeIfNotExists.go
```diff
@@ -32,63 +33,54 @@ func (np nodeProvider) CreateNodeIfNotExists(ctx context.Context) (node.Node, er
...
+		np.config.DataDir,
+		"--api-listen-address",
+		np.config.APIListenAddress,
 		"--container-runtime",
 		np.config.ContainerRuntime,
 		"node",
 		"create",
+	}
+
+	if os.Geteuid() != 0 {
```
chrisdostert, line 48: opctl must run as root now in order to manage dns and network

sdks/go/node/api/listen.go
```diff
@@ -73,31 +45,23 @@ func (hd _httpListener) listen(
...
 	router.PathPrefix("/").Handler(http.FileServer(http.FS(buildFS)))
 
-	httpServer := http.Server{
+	apiServer := http.Server{
 		Addr:    address,
 		Handler: handlers.CORS()(router),
 	}
 
-	// catch signals to ensure shutdown properly happens
-	done := make(chan os.Signal, 1)
```
chrisdostert, line 82: pulled this out of the listener since we have multiple listeners now (API and DNS); we want to capture signal above them and handle via context.

sdks/go/node/api/urltemplates/urlTemplates.go
```diff
@@ -0,0 +1,11 @@
+package urltemplates
+
+/* resources */
```
chrisdostert, line 3: moving this was required after moving `Listen` out of `cli` to `sdks/go/node/api` to avoid import cycles

rajivbaviskar (approved):

cli/internal/nodeprovider/local/createNodeIfNotExists.go
zachpuck: Have we considered other options like only running the dns configuration steps as root in an isolated way. if a user doesnt want to run as root are they unable to use this version?

chrisdostert: It's effectively already in an isolated way in the sense It's only the node/daemon that's required to run as root. This is similar to how other tools do it e.g. docker for Mac takes this same approach.

rajivbaviskar (approved):

PR #1113, opened 2024-11-07
Ensure permissions on opctl data can't be broken by umask

chrisdostert:
This PR fixes an edge case encountered in testing on OSX following #1112 where variables were initialized by mounting to container files/dirs would error with `invalid mount config [...] permission denied`.

Upon investigation it was found that:
On Docker for Mac, the file sync between VM and host doesn't have access to `root` (UID=0) owned files unless the owning group also has access. 

Additionally, by default umask is `022` so this was leading to no owning group write access and the error scenario.

### DIFF
--- a/cli/nodeCreate.go
+++ b/cli/nodeCreate.go
@@ -57,14 +57,19 @@ func nodeCreate(
 				cliColorer.Info(fmt.Sprintf("opctl API listening at %s", nodeConfig.APIListenAddress)),
 			)
 
+			c, err := core.New(
+				ctx,
+				containerRT,
+				dataDir.Path(),
+			)
+			if err != nil {
+				return err
+			}
+
 			return api.Listen(
 				ctx,
 				nodeConfig.APIListenAddress,
-				core.New(
-					ctx,
-					containerRT,
-					dataDir.Path(),
-				),
+				c,
 			)
 		},
 	)
--- a/sdks/go/node/datadir/datadir.go
+++ b/sdks/go/node/datadir/datadir.go
@@ -6,6 +6,7 @@ import (
 	"path/filepath"
 
 	"github.com/rogpeppe/go-internal/lockedfile"
+	"golang.org/x/sys/unix"
 )
 
 // DataDir is an interface exposing the functionality we require in conjunction with our "data dir".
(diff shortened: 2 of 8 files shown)


cli/nodeCreate.go
```diff
@@ -57,14 +57,19 @@ func nodeCreate(
 				cliColorer.Info(fmt.Sprintf("opctl API listening at %s", nodeConfig.APIListenAddress)),
 			)
 
+			c, err := core.New(
```
chrisdostert, line 60: unrelated; touch up to pass through errors creating node

chris-spalding:
> Upon investigation it was found that:
> On Docker for Mac, the file sync between VM and host doesn't have access to root (UID=0) owned files unless the owning group also has access.
> Additionally, by default umask is 022 so this was leading to no owning group write access and the error scenario.

Great find! How did you figure out the issue and the default umask?

sdks/go/node/datadir/datadir.go
```diff
@@ -21,10 +22,10 @@ type DataDir interface {
 func ensureExists(
 	resolvedDataDirPath string,
 ) error {
-	if err := os.MkdirAll(resolvedDataDirPath, 0775|os.ModeSetgid); err != nil {
-		return err
-	}
-	return nil
+	// don't de-privilege group
+	unix.Umask(0002)
```
chris-spalding, line 26: OOC - would the following have the same effect? Or would it prevent the Docker VM (or some other user/group) from being able to read the file?
```suggestion
	unix.Umask(0000)
```

chris-spalding (approved):
LGTM! I had some question for my own edification but I understand what this does and why it works. Really good find - the issue seems quite hard to track down

sdks/go/node/datadir/datadir.go
chrisdostert: this would work as well it just is relaxed beyond what we need; it would also remove the mask from "other"  and we only need it removed for "group"

chris-spalding: Ahh I see - I thought the `0` mask in the 'other' position would remove any 'other' perms but it sounds a `0` just allows any existing perms to persist. Thanks for clarifying!

PR #1114, opened 2024-11-08
Don't create files as root

chrisdostert:
Upon further testing, Docker for Mac file sync between the host and the VM runs as an unprivileged user. This means it has no permissions to read or write root owned files from the host. Following #1112 , this causes issues instantiating variables from container files/dirs since behind the scenes we create files/dirs and since we now run as root they're created as root. 

This PR addresses this issue by introducing an `unsudo` package that will chown any files/dirs opctl to the `uid` and `gid` of the user that started the opctl node.


website/docs/training/containers/container-networking.md
```diff
@@ -16,7 +16,7 @@ If multiple containers have the same [name](../../reference/opspec/op-directory/
     run:
       parallel:
         - container:
-            image: { ref: alpine }
+            image: { ref: nginx }
```
chrisdostert, line 19: unrelated) fix for training article

.markdownlint/customRule.js
```diff
@@ -9,13 +9,13 @@ module.exports = {
...
-      if (heading.tag === "h1" && heading.line !== "# Change Log") {
+      if (heading.tag === "h1" && heading.line !== "# Changelog") {
         // this explicit text is the only valid content for the first heading
         return onError({
           "lineNumber": heading.lineNumber,
-          "detail": "First heading should be '# Change Log'.",
+          "detail": "First heading should be '# Changelog'.",
         });
-      } else if (heading.tag === "h2" && !/^## \d\.\d{1,9}\.\d{1,3}(-[.A-Za-z0-9]+)? - \d{4}-\d{2}-\d{2}$/.test(heading.line)) {
+      } else if (heading.tag === "h2" && !/^## \[(\d\.\d{1,9}\.\d{1,3}(-[.A-Za-z0-9]+)?)|Unreleased\]| - \d{4}-\d{2}-\d{2}$/.test(heading.line)) {
```
chrisdostert, line 18: As it turns out, keepachangelog requires surrounding version by brackets i.e. `[semver]` and also supports `Unreleased` as a version to designate work in progress not yet ready to release.

.opspec/release/to-github/op.yml
```diff
@@ -28,31 +33,18 @@ run:
...
-            else
-              echo -n false > /prerelease
-            fi
-        files:
-          /prerelease: $(prerelease)
     - op:
         ref: github.com/opspec-pkgs/github.release.create#2.0.0
         inputs:
           commitish: $(commit)
+          description: $(latestRelease.description)
```
chrisdostert, line 40: use changelog details as release description so user facing product changes are presented as opposed to GitHub default of summary of commits (which give many irrelevant non-product facing changes).

sdks/go/node/dns/listen.go
```diff
@@ -24,8 +25,11 @@ func Listen(
 
 	go func() {
 		<-ctx.Done()
+		ctx, cancel := context.WithTimeout(ctx, 5*time.Second)
+		defer cancel()
+
 		// little hammer
-		dnsServer.Shutdown()
+		dnsServer.ShutdownContext(ctx)
```
chrisdostert, line 32: unrelated touch up to limit how long we wait for dns listener to shutdown

rajivbaviskar (approved):

rajivbaviskar (approved):

PR #1117, opened 2024-11-11
Release prep; Fix container dns names on linux & support passwordless sudo for node startup

chrisdostert:
This PR addresses two last touchups before release 0.1.56 namely:
- not requiring explicitly using sudo to start a node if the caller has passwordless sudo enabled
- address a regression for the linux os where container dns names wouldn't work

### DIFF
--- a/.opspec/changelog/getLatestRelease/main.go
+++ b/.opspec/changelog/getLatestRelease/main.go
@@ -2,8 +2,9 @@ package main
 
 import (
 	"encoding/json"
-	"golang.org/x/mod/semver"
 	"os"
+
+	"github.com/Masterminds/semver/v3"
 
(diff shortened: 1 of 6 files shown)


.opspec/changelog/getLatestRelease/main.go
```diff
@@ -2,9 +2,10 @@ package main
 
 import (
 	"encoding/json"
-	"golang.org/x/mod/semver"
 	"os"
 
+	"github.com/Masterminds/semver/v3"
```
chrisdostert, line 7: unrelated but turns out `golang.org/x/mod/semver` doesn't handle pre releases correctly so resolving here.

chris-spalding (approved):

harshita-chaudhary (approved):

PR #1118, opened 2024-11-12
Expose rate limiting errors from gpu enablement check

chrisdostert:
This PR passes through an error condition that was being ignored while testing if the machine is gpu enabled. Generally the only way this error occurs is during rate limiting when the image is not cached, which is a rather unlikely edge case, but are hitting it ourselves in e2e tests due to no not caching any images across test scenarios.

### DIFF
--- a/cli/.opspec/test/e2e/op.yml
+++ b/cli/.opspec/test/e2e/op.yml
@@ -38,4 +38,4 @@ run:
                   /cmd.sh:
                   /usr/local/bin/opctl: $(../../../opctl-linux-amd64)
                 image:
-                  ref: docker:20.10.17-dind
+                  ref: docker:27.3-dind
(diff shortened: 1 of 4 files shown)


cli/.opspec/test/e2e/op.yml
```diff
@@ -37,5 +41,7 @@ run:
                   /args.yml: $(scenario.call.scope)
                   /cmd.sh:
                   /usr/local/bin/opctl: $(../../../opctl-linux-amd64)
+                sockets:
+                  /var/run/docker.sock: $(dockerSocket)
```
chrisdostert, line 45: this adds image caching (via reusing the same docker daemon across all tests)

sdks/go/node/containerruntime/docker/isGpuSupported.go
```diff
@@ -54,6 +54,9 @@ func isGpuSupported(
 		"",
 		noOpEventPublisher{},
 	)
+	if err != nil {
```
chrisdostert, line 57: this is the actual fix for the lack of rate limiting error surfacing during gpu testing

rajivbaviskar (approved):

PR #1120, opened 2024-11-14
Fix executable bit not maintained within remote ops

chrisdostert:
This PR addresses a regression introduced in 0.1.56, specifically that for remote ops containing files relying on execution bit being set, failed with permission errors due to the execution bit not being maintained.


.github/workflows/build.yml
```diff
@@ -17,7 +17,7 @@ jobs:
       with:
         fetch-depth: 0
     - name: Install Opctl
-      run: curl -L https://github.com/opctl/opctl/releases/download/v0.1.55/opctl-linux-amd64.tgz | sudo tar -xzv -C /usr/local/bin
+      run: curl -L https://github.com/opctl/opctl/releases/latest/download/opctl-linux-amd64.tgz | sudo tar -xzv -C /usr/local/bin
```
chrisdostert, line 20: these are unrelated but just set us back to using latest opctl now that bugs are addressed.

.opspec/compile/op.yml
```diff
@@ -0,0 +1,19 @@
+name: build
```
scottb-remitly: this op is in the `compile` dir but named `build` - could it be made consistent

scottb-remitly (approved):
Looks good, love the added unit test 👍

PR #1122, opened 2024-11-17
Fix release notes with JSON special characters break release

chrisdostert:
This is PR fixes an issue where if our change log release notes included JSON special characters such as double quotes, it would break the release process. This was because the `github.com/opspec-pkgs/github.release.create` op was not properly encoding them. This issue has been resolved in the latest version of that op so updating to it will resolve it. 

Unrelated:
We also updated to latest version of `github.com/opspec-pkgs/github.release.upload` in order to only need to pass `accessToken` instead of username & accessToken

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -17,7 +17,7 @@ jobs:
       with:
         fetch-depth: 0
     - name: Install Opctl
-      run: curl -L https://github.com/opctl/opctl/releases/download/v0.1.55/opctl-linux-amd64.tgz | sudo tar -xzv -C /usr/local/bin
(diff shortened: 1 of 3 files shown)


.github/workflows/build.yml
```diff
@@ -17,7 +17,7 @@ jobs:
       with:
         fetch-depth: 0
     - name: Install Opctl
-      run: curl -L https://github.com/opctl/opctl/releases/download/v0.1.55/opctl-linux-amd64.tgz | sudo tar -xzv -C /usr/local/bin
+      run: curl -L https://github.com/opctl/opctl/releases/latest/download/opctl-linux-amd64.tgz | sudo tar -xzv -C /usr/local/bin
```
chrisdostert, line 20: unrelated, use latest opctl now that reported issues resolved

zachpuck (approved):

chris-spalding (approved):

PR #1123, opened 2024-11-17
Fix 'No such container: opctl_[...]' errors if container exits too fast (race condition)

chrisdostert:
This PR fixes an error `No such container: opctl_[...]` which could occur if containers exit before we can inspect ip, stream logs, or status code. This was a race condition effecting docker container runtime caused by usage of `AutoRemove: true` in the containers `HostConfig`.

Unrelated: I also upgraded the docker SDK to latest while debugging and although it wasn't the issue I kept it because being on latest version is worthwhile anyways.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -17,7 +17,7 @@ jobs:
       with:
         fetch-depth: 0
     - name: Install Opctl
-      run: curl -L https://github.com/opctl/opctl/releases/latest/download/opctl-linux-amd64.tgz | sudo tar -xzv -C /usr/local/bin
+      run: curl -L https://github.com/opctl/opctl/releases/download/v0.1.55/opctl-linux-amd64.tgz | sudo tar -xzv -C /usr/local/bin
     - name: Check for CHANGELOG.md
       run: opctl run changelog/find-in-diff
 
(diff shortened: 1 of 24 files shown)


sdks/go/node/containerruntime/docker/runContainer.go
```diff
@@ -190,49 +185,49 @@ func (cr _runContainer) RunContainer(
...
-	}
+	eg, ctx := errgroup.WithContext(ctx)
 
-	if endpointSettings, ok := containerJSON.NetworkSettings.Networks[networkName]; ok && req.Name != nil {
-		err = dns.RegisterName(
-			ctx,
-			*req.Name,
-			endpointSettings.IPAddress,
-		)
+	eg.Go(
```
chrisdostert, line 190: This was unrelated but just cleans up the code to be more readable & less complex.

sdks/go/internal/unsudo/createDir.go
```diff
@@ -18,16 +18,17 @@ func CreateDir(
 	for i := 0; i < len(parts); i++ {
 		currentPath = filepath.Join(currentPath, parts[i])
 
-		// Check if the directory already exists
-		_, err := os.Stat(currentPath)
-		if os.IsNotExist(err) {
-			if err := os.Mkdir(currentPath, 0700); err != nil {
-				return err
+		if err := os.Mkdir(currentPath, 0700); err != nil {
```
chrisdostert, line 21: always try to create to avoid races

.github/workflows/build.yml
```diff
@@ -17,7 +17,7 @@ jobs:
       with:
         fetch-depth: 0
     - name: Install Opctl
-      run: curl -L https://github.com/opctl/opctl/releases/latest/download/opctl-linux-amd64.tgz | sudo tar -xzv -C /usr/local/bin
+      run: curl -L https://github.com/opctl/opctl/releases/download/v0.1.55/opctl-linux-amd64.tgz | sudo tar -xzv -C /usr/local/bin
```
chrisdostert, line 20: need to drop to 0.1.55 in order to release this because we're hitting the issue this PR is fixing ; )

rajivbaviskar (approved):

rajivbaviskar (approved):

PR #1140, opened 2025-02-08
Native privilege escalation

chrisdostert:
This PR introduces a new experience for privilege escalation.

The current user experience when opctl needed elevated privileges, it would simply error out and tell the user to call it with sudo. 

This presented a bad user experience for a few reasons:
- it requires users to re-run the command with sudo (which is annoying)
- it requires users know & remember the intricacies of sudo usage such as passing the `-E` flag to maintain the users current environment (which is error prone)

The user experience introduced in this PR is that when opctl needs elevated privileges, it will simply ask the user for their password and then proceed to run their original command. 

Thanks @zachpuck for catching this bad user experience and pushing for the improvement!


cli/internal/nodeprovider/local/createNodeIfNotExists.go
```diff
@@ -26,19 +26,15 @@ func (np nodeProvider) CreateNodeIfNotExists(
...
-		ctx,
-		np.config.DataDir,
-	)
-	if err != nil {
-		return nil, err
-	}
-
-	if nodeProcess != nil {
+	// check if node API reachable
+	if err := apiClientNode.Liveness(ctx); err == nil {
```
chrisdostert, line 30: replacing pidfile.TryGetProcess with apiClientNode.Liveness because the former only works as root and otherwise returns nothing whereas the latter works whether root or not.

rajivbaviskar (approved):

Issue #1137, opened 2025-02-05
Passing `container.image.pullCreds` For Registry Other Than `docker.io` On Linux Causes Failure


squatched:
## 🐛 Bug Report

When passing credentials to a `container.image.pullCreds` step on a Linux host for an image hosted outside of `docker.io`, you get the following failure:
```
<your image hosted somewhere else, say AWS ECR> crashed: Error response from daemon: Head "https://registry-1.docker.io/v2/library/alpine/manifests/latest": unauthorized: incorrect username or password
```

## To Reproduce

(Write your steps here:)

1. Get access to an image hosted somewhere that requires credentials outside of `docker.io`.
1. Setup an op to use that image.
1. `run` the op under Linux.

Sample op:
```yml
name: test-pullCreds
description: Test pullCreds for image.

inputs:
  awsImage:
    string:
      description: The image to try and pull and run a command in.
  HOME:
    dir:
      description: Home directory (needed for AWS creds)
  dockerSocket:
    socket:
      description: socket of the host machine's docker

run:
  serial:
    # First auth with AWS/ECR.
    - container:
        image: { ref: amazon/aws-cli }
        cmd:
          - sh
          - -ce
          - |
            aws --region=us-west-2 ecr get-login-password  >/pullCredsPassword
        dirs:
          /root/.aws: $(HOME/.aws)
        files:
          /pullCredsPassword: $(pullCredsPassword)
    # Try to use them via opctl pullCreds.
    - container:
        image:
          ref: $(awsImage)
          pullCreds:
            username: AWS
            password: $(pullCredsPassword)
        cmd:
          - sh
          - -ce
          - printf "********** Success! **********\n"

```

## Expected behavior

For the creds specified to be used only to pull the specified image.

## Actual Behavior

Opctl tries to pull `alpine` from `docker.io` with passed `pullCreds` values (which don't work, naturally).

The problematic behavior is defined in https://github.com/opctl/opctl/blob/main/sdks/go/node/containerruntime/docker/isGpuSupported.go

PR #1141, opened 2025-02-08
Ghcr alpine

chrisdostert:
This PR moves the codebase away from using dockerhub to using github container registry for any alpine images since alpine now officially publish there, which aside from being one less service dependency (we already depend on github), has no rate limiting.

closes https://github.com/opctl/opctl/issues/1137


zachpuck (approved):
Needs a changelog update but otherwise LGTM

chrisdostert:
> Needs a changelog update but otherwise LGTM

Thanks, that check is optional; I feel like we should make it less scary i.e. maybe not fail but just show info or something.

Either way, in this case I wasn't intending to create a release as It's largely just changing the code experience and not product level.

PR #1146, opened 2025-02-10
Automatic handling of mDNSResponder port conflict on OSX

chrisdostert:
This PR
- automates the handling of an edge case where opctl cannot start on OSX due to a port conflict caused by MDNSResponder which binds to `*:53` i.e. port 53 on all IPs & all adapters.
- fixes binding to a non-standard dns port (e.g. `54`) in the case the current host OS supports it (i.e. OSX).


.opspec/release/to-ghcr/dood.Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM --platform=$TARGETPLATFORM ghcr.io/linuxcontainers/alpine:3.11.3
+FROM --platform=$TARGETPLATFORM ghcr.io/linuxcontainers/alpine:3.18
```
chrisdostert, line 1: Unrelated; this is fixing the fact that this image isn't available in Github Container Registry (only identified via last build failing)

cli/internal/pidfile/tryGetProcess.go
```diff
@@ -6,7 +6,7 @@ import (
 	"fmt"
 	"os"
 
-	"github.com/shirou/gopsutil/v3/process"
+	"github.com/shirou/gopsutil/v4/process"
```
chrisdostert, line 9: not technically required but updating to v4 while we're at it.

rajivbaviskar (approved):

zachpuck (approved):

Issue #744, opened 2020-04-23
Default pkg source credentials


apexskier:
## 🚀 Feature

Add support for configurable or "default" pkg source credentials, to avoid needing to re-enter on the command line each time and make it easier to automate ops in a non-interactive environment.

This could be implemented in a number of ways. My proposal is to storing some config yml file in `$XDG_CONFIG_HOME/opctl/credentials.yml` with a way to match package sources with credentials.

The structure could be something like:

```yml
credentials:
- source: github.com/Remitly
  username: $(github.username) # environment variable / object reference
  password: $(github.accessToken)
- source: github.com
  username: myusername # plain text
  password: $(GITHUB_PASSWORD) # environment variable
```

This attempts to math opctl's opspec format. It uses opctl's [object property referencing](https://opctl.io/docs/opspec/reference/types/object#property-referencing) syntax and its ability to [read environment variables](https://opctl.io/docs/zero-to-hero/inputs-outputs).

It would use the credential source to match the pkg source's prefix and use the first credential found (e.g. `opctl run -a apiKey=$LOKALISE_TOKEN github.com/Remitly/parakeet#v1.0.0/parakeet-ops/update-all-locales` would match the first credential, not the second).

chrisdostert:
@apexskier thanks for filing!

Just wanna clarify, is the intent only to use these creds to pull the root op of a call (github.com/Remitly/parakeet#v1.0.0/parakeet-ops/update-all-locales in your example) i.e. not calls within the call graph thereafter?

chrisdostert:
Also, per https://github.com/opctl/opctl/issues/745 just thought i'd mention, another possibility could be to allow git cred store to be a default source

chrisdostert:
Also worth mentioning, pkgs do get cached; you shouldn't be prompted more than once per pkg, but I know that doesn't solve the underlying thing you're after.

chrisdostert:
There's been the idea of a `space` floating around for some time. Idea was to solve for at least configuration of op inputs across all the things. So your ops would be started within a given space and therefore have access to variables defined in said space. It would/could replace things like credential stores and env vars. It's particularly useful to tie together the experience from the CLI and the UI; obviously the UI doesn't have access to things like env vars.

apexskier:
[...]

PR #1157, opened 2025-04-01
Support default ops for git repositories & default `pullCreds` for ops

chrisdostert:
This PR adds the following two related capabilities to opctl:
- Support default ops within git repositories via remote conventional `.opctl` git repository
- Support default `pullCreds` for ops via `opctl auth [...]`

Starting with this version, when you `opctl ls` within a git repository, opctl will check if a conventional `.opctl` repository exists at the remote upstream. If it does, it will try to run the `getConfig` op and any ops returned as an output will be:
- shown by `opctl ls`
- in scope such that they can be `opctl run [...]`.

If auth is required to check for a remote `.opctl` repository, opctl will provide helpful instructions e.g.
```
run "opctl auth add github.com -u <username> -p <password>" to complete setup

REF     DESCRIPTION
[...]
```

closes https://github.com/opctl/opctl/issues/744


chrisdostert:
breaking up into 2 PRs

PR #1159, opened 2025-04-02
Support specifying image.platform.arch in container calls

chrisdostert:
This PR adds the ability to specify image platform architecture (e.g. `amd64`) in order to ensure containers requiring images built for a particular architecture don't break when run on nodes with different architecture.

example op using it:
```yaml
name: examplePlatformUsage
run:
  container:
    image:
      ref: amd64/python:3.12
      platform:
        arch: amd64
```

Unrelated:
This PR also removes container.image.src attribute in the opspec JSON schema; it is [not used/supported in code](https://github.com/opctl/opctl/blob/155cc1f46837790c9da3be8c7b1f346ca07d503a/sdks/go/model/opSpec.go#L46)

closes #1156

### DIFF
--- a/opspec/opfile/jsonschema.json
+++ b/opspec/opfile/jsonschema.json
@@ -926,12 +926,19 @@
                   "description": "Image reference to resolve from network.",
                   "$ref": "#/definitions/expression"
                 },
+                "platform": {
+                  "type": "object",
+                  "properties": {
+                    "arch": {
+                      "description": "Architecture of image. MUST be a valid [v1.0.1 OCI (Open Container Initiative) `image-index`](https://github.com/opencontainers/image-spec/blob/v1.0.1/image-index.md) platform architecture"
+                    }
+                  },
+                  "required": [
+                    "arch"
+                  ]
+                },
                 "pullCreds": {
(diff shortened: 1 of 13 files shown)


opspec/opfile/jsonschema.json
```diff
@@ -926,12 +926,19 @@
...
+                  "required": [
+                    "arch"
+                  ]
+                },
                 "pullCreds": {
                   "$ref": "#/definitions/pullCreds"
-                },
-                "src": {
-                  "description": "Source of image. MUST be a valid [v1.0.1 OCI (Open Container Initiative) `image-layout`](https://github.com/opencontainers/image-spec/blob/v1.0.1/image-layout.md) directory.",
-                  "type": "string"
```
chris-plucker, line 934: Looks like `src` is still supported? Do we want to remove this from the jsonschema?

chrisdostert: Good call out; I'll add this to the description. This was an unrelated "cleanup" because [I noticed that it technically isn't used/supported](https://github.com/opctl/opctl/blob/155cc1f46837790c9da3be8c7b1f346ca07d503a/sdks/go/model/opSpec.go#L46)

PR #1160, opened 2025-04-03
Support default auth for pulling ops via `opctl auth [...]`

### LINKED ISSUE #808: Stop using Fake pubsub in tests; use real pubsub.

chrisdostert:
Before this PR you could use `opctl auth add [...]` to add default auth for pulling images.

After this PR you can also use it to add default auth for pulling ops.

Note: 
in both cases this default auth can still be overridden by explicitly providing pullCreds in you container or op call. 

closes #808 #744


zachpuck (dismissed):
Nice enhancement 🚀

chrisdostert:
> Any test cases added for this? Only test changes I saw were structural given the new state store being passed around. At a minimum, I would expect:
> 
> * Given an op without `pullCreds` but defaults specified, are the defaults used?
> * Given an op with `pullCreds` and defaults specified, are the op `pullCreds` passed?

Added some e2e tests to cover both success and negative cases as well as calling private ops directly or nested within other ops.

PR #1163, opened 2025-04-06
Migrate from mow.cli to cobra

### LINKED ISSUE #992: Migrate CLI to https://github.com/spf13/cobra
## 💥 Proposal

Currently we use github.com/jawher/mow.cli as our CLI framework.

this comes with a few shortcomings that would be nice to overcome; namely:
- no "did you mean [similar command]" when someone typos the command
- no man page generation
- no shell completion generation
- no arbitrary option placement; mow.cli requires you specify a spec for each command, inclusive of opts and args. It doesn't support providing the opts and args out of order. This is sortof a non-ideal UX. 
- not as actively maintained or popular as other go CLI frameworks like cobra
- [hardcoded](https://github.com/jawher/mow.cli/blob/d9d0f2e822e1fcc4b33e338f2181daafcb4d9077/cli.go#L89) to print version info to stderr as opposed to stdout

It's proposed we move to [https://github.com/spf13/cobra](https://github.com/spf13/cobra) which gives us all of the above.

### LINKED ISSUE #1162: `opctl --version` to `stdout`
## 💥 Proposal

Currently, `opctl --version` sends its output to `stderr`. Can it be be sent to `stdout` instead? It will be helpful for scripting purposes (e.g., checking if version in the environment matches a requirement).

chrisdostert:
This PR:
- migrates the opctl CLI from mow.cli to cobra; see #992 for reasoning
- deletes the manually maintained CLI reference docs in favor of autogenerated docs (using the cobra CLI markdown generation feature)
- fixes an edge case where if opctl wasn't on the callers PATH it might error with `sudo: opctl command not found`

closes #1162
closes #992


cli/internal/euid0/ensure.go
```diff
@@ -3,18 +3,29 @@ package euid0
...
 // Ensure EUID is 0 or return error
 func Ensure() error {
 	if os.Geteuid() != 0 {
-		cmd := exec.Command("sudo", append([]string{"-E", os.Args[0]}, os.Args[1:]...)...)
+		pathToOpctlBin, err := os.Executable()
+		if err != nil {
+			return err
+		}
+
+		pathToOpctlBin, err = filepath.EvalSymlinks(pathToOpctlBin)
```
rajivbaviskar, line 17: Is this for `fixes an edge case where if opctl wasn't on the callers PATH it might error with sudo: opctl command not found` ?

rajivbaviskar (approved):

cli/internal/euid0/ensure.go
rajivbaviskar: Chatted offline on this.

PR #1174, opened 2026-03-19
Detect re-tagged git ops and re-fetch when the tag has moved

gedi-remitly:
## Background

opctl caches git ops locally and uses the presence of a marker file to skip re-cloning on subsequent runs. This works fine for immutable tags, but git tags are mutable — they can be force-pushed to point at a different commit at any time.

A pattern common in the GitHub Actions ecosystem (and [recommended by GitHub](https://docs.github.com/en/actions/how-tos/create-and-publish-actions/using-immutable-releases-and-tags-to-manage-your-actions-releases)) is to maintain a "floating" major version tag (e.g. `v1`) that always points to the latest stable release within that major. This lets consumers pin to `myop#v1` and receive updates automatically without changing their op reference.

opctl's current caching strategy makes this impossible — once `v1` is cached, it is never re-fetched regardless of where the tag now points.

## What this changes

Before cloning (or deciding to skip a clone), opctl now runs the equivalent of `git ls-remote` against the remote to resolve what commit the tag currently points to. It compares that hash against what was cached locally:

- If the hashes match, the cache is used as before.
- If the hashes differ (tag was re-pointed), the op is re-cloned into a temporary path and atomically replaces the cached copy only after the clone succeeds, so a failed re-clone never corrupts the existing cache.

### DIFF
--- a/.opspec/changelog/find-in-diff/op.yml
+++ b/.opspec/changelog/find-in-diff/op.yml
@@ -3,7 +3,7 @@ name: check-for-changelog
 run:
   container:
     image:
-      ref: bitnami/git:2.45.0
+      ref: bitnami/git:latest
     cmd:
       - /bin/sh
       - -c
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file in
 accordance with
 [![keepachangelog 1.0.0](https://img.shields.io/badge/keepachangelog-1.0.0-brightgreen.svg)](http://keepachangelog.com/en/1.0.0/)
 
+## [0.1.75] - 2026-03-19
+
+### Fixed
+
+- Git ops with re-pointed tags (e.g. floating `v1` tags) are now re-fetched when the remote tag points to a new commit
+
 ## [0.1.74] - 2025-09-26
 
 ### Fixed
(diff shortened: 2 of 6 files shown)


rajivbaviskar (dismissed):
LGTM

rajivbaviskar (dismissed):

chrisdostert:
Just noting, the existing behavior was by design. I get what you're after, which is preventing the need to bump a bunch of consumer op references to roll out a change, but that behavior is risky.  There's no guarantees of backwards compatibility so allowing re-publishing introduces the potential of breaking anyone/everyone depending on your op without them ever changing anything. 

The idea you as a consumer change nothing and things might break seems less than ideal. Not sure if you've considered other ideas. 

Just my 2c.

gedi-remitly:
Moved these changes to https://github.com/opctl/opctl/pull/1175 because this PR needed approvals to run CI due to it being based of a fork.

@chrisdostert I agree that pointing a Git tag to another commit can be risky. But opctl cannot prevent that. If opspec maintainers force-push a tag and opctl does not check for the commit the tag is pointing to, some user will get the newest commit while others will continue using the stale one. This change adjusts opctl's caching algorithm to account for the reality that Git tags can be force-pushed. It does not endorse usage of floating (mutable) Git tags.

PR #1175, opened 2026-03-23
Detect re-tagged git ops and re-fetch when the tag has moved

gedi-remitly:
## Background

opctl caches git ops locally and uses the presence of a marker file to skip re-cloning on subsequent runs. This works fine for immutable tags, but git tags are mutable — they can be force-pushed to point at a different commit at any time.

A pattern common in the GitHub Actions ecosystem (and [recommended by GitHub](https://docs.github.com/en/actions/how-tos/create-and-publish-actions/using-immutable-releases-and-tags-to-manage-your-actions-releases)) is to maintain a "floating" major version tag (e.g. `v1`) that always points to the latest stable release within that major. This lets consumers pin to `myop#v1` and receive updates automatically without changing their op reference.

opctl's current caching strategy makes this impossible — once `v1` is cached, it is never re-fetched regardless of where the tag now points.

## What this changes

Before cloning (or deciding to skip a clone), opctl now runs the equivalent of `git ls-remote` against the remote to resolve what commit the tag currently points to. It compares that hash against what was cached locally:

- If the hashes match, the cache is used as before.
- If the hashes differ (tag was re-pointed), the op is re-cloned into a temporary path and atomically replaces the cached copy only after the clone succeeds, so a failed re-clone never corrupts the existing cache.

### DIFF
--- a/.opspec/changelog/find-in-diff/op.yml
+++ b/.opspec/changelog/find-in-diff/op.yml
@@ -3,7 +3,7 @@ name: check-for-changelog
 run:
   container:
     image:
-      ref: bitnami/git:2.45.0
+      ref: bitnami/git:latest
     cmd:
       - /bin/sh
       - -c
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file in
 accordance with
 [![keepachangelog 1.0.0](https://img.shields.io/badge/keepachangelog-1.0.0-brightgreen.svg)](http://keepachangelog.com/en/1.0.0/)
 
+## [0.1.75] - 2026-03-19
+
+### Fixed
+
+- Git ops with re-pointed tags (e.g. floating `v1` tags) are now re-fetched when the remote tag points to a new commit
+
 ## [0.1.74] - 2025-09-26
 
 ### Fixed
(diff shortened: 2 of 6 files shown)


.opspec/changelog/find-in-diff/op.yml
```diff
@@ -3,7 +3,7 @@ name: check-for-changelog
 run:
   container:
     image:
-      ref: bitnami/git:2.45.0
+      ref: bitnami/git:latest
```
chris-plucker, line 6: Ironically, do we want to pin to `latest` on this?

chris-plucker (approved):
Makes sense to me.

Regarding Chris's comment on #1174:
> The idea you as a consumer change nothing and things might break seems less than ideal. Not sure if you've considered other ideas.

I think this is fairly transparent to users if they are setting a `v1` reference then they are likely expecting to be on the latest of the `v1.x.x` version. If they don't want that behavior, they should tag with a fully qualified version.

.opspec/changelog/find-in-diff/op.yml
gedi-remitly: Bitnami stopped publishing tags for their Docker images (except for latest) unless you subscribe to their paid plan. I could not find a sufficiently trustworthy alternative to bitnami/git on Docker Hub.

PR #1176, opened 2026-03-23
CI check: verify op install test on latest main

gedi-remitly:
Temporary PR to verify whether the `cli > op > install > should not err` test failure is pre-existing on main or introduced by #1175.


gedi-remitly:
Build passed — confirmed the op install test failure is introduced by #1175, not pre-existing on main.

PR #1186, opened 2026-06-18
Propagate proxy env vars from the opctl node into op containers

gedi-remitly:
## Problem

opctl runs ops as containers via the container runtime's SDK (e.g. the Docker Go SDK), not the `docker` CLI. On hosts whose only egress route to the internet is an HTTP/HTTPS **forward proxy** (e.g. AWS CodeBuild runners, or any CI behind a corporate proxy, with no direct/NAT route), the processes inside op containers — `go mod download`, `apt-get`, `curl`, etc. — need `HTTP_PROXY` / `HTTPS_PROXY` / `NO_PROXY` / `ALL_PROXY` in their environment to reach the network.

Today opctl does not propagate these, so those containers run with no proxy configuration and their network calls hang or fail.

Because opctl talks to the daemon via the SDK rather than the CLI, the usual `~/.docker/config.json` `proxies` convenience does **not** apply here — opctl has to inject the variables itself.

(The image *pull* is a separate, daemon-side concern and is intentionally out of scope.)

## Change

Add a small shared helper `containerruntime.ProxyEnvVars` that, given the op-provided env map, returns the proxy related variables present in the **opctl node's own process environment** that the op did **not** already set:

- Variables propagated: `HTTP_PROXY`, `HTTPS_PROXY`, `NO_PROXY`, `ALL_PROXY` and their lowercase forms.
- Non-override rule: if the op explicitly set a key, the node value is **never** added/clobbered for that key.
- When no proxy vars are present in the node env, the container env is unchanged.

It's wired into both runtimes:

- **Docker**: `sdks/go/node/containerruntime/docker/constructContainerConfig.go`
- **Kubernetes**: `sdks/go/node/containerruntime/k8s/constructPod.go`

## Tests

- `sdks/go/node/containerruntime/proxyEnv_test.go` (`TestProxyEnvVars`) covers: (a) node-env proxy vars are propagated, (b) op-provided values are not overridden, (c) no proxy vars => empty result. Uses save/restore of env so the host environment can't perturb results.
- An added Ginkgo spec in `constructContainerConfig_test.go` asserts propagation flows into `container.Config.Env` and that an op-set `HTTP_PROXY` wins. The existing exact-match spec was made deterministic by clearing proxy vars for its duration.

`go build ./node/containerruntime/...` and `go test ./node/containerruntime/` pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,13 @@ All notable changes to this project will be documented in this file in
 accordance with
 [![keepachangelog 1.0.0](https://img.shields.io/badge/keepachangelog-1.0.0-brightgreen.svg)](http://keepachangelog.com/en/1.0.0/)
 
+## [0.1.76] - 2026-06-18
+
+### Added
+
+- Proxy env vars (`HTTP_PROXY`, `HTTPS_PROXY`, `NO_PROXY`, `ALL_PROXY`, and their lowercase forms) from the
+opctl node's environment are now propagated into op containers, unless the op already set them
+
 ## [0.1.75] - 2026-03-19
 
 ### Fixed
--- a/cli/internal/nodeprovider/local/createNodeIfNotExists.go
+++ b/cli/internal/nodeprovider/local/createNodeIfNotExists.go
@@ -16,6 +16,7 @@ import (
 
 	"github.com/opctl/opctl/cli/internal/euid0"
 	"github.com/opctl/opctl/sdks/go/node"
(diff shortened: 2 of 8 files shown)


gedi-remitly:
Superseded by #1187, which carries the identical change plus a CHANGELOG entry, with the head branch hosted in `opctl/opctl` instead of a fork. That's required because GitHub does not expose repo secrets (here `TEST_GITHUB_ACCESS_TOKEN`) to pull requests from forks, so the `Build` workflow's `test` step can't pass on this fork-based PR. See #1187 for the green `Build`.

